Issue #12, opened 2018-11-29
DefaultRepairConfigurationProvider not handling replication changes properly


emolsson:
When **#onKeyspaceChanged()** is called it checks with _ReplicatedTableProvider_ if the keyspace is replicated enough for repair.

In the case when a keyspace changes replication the token metadata is not properly updated until **after** the listeners have been notified of the keyspace update. This means that going from 1 -> 3 in rf won't cause the table to be scheduled for repair. But changing rf from 3 -> 1 will schedule it.

PR #17, opened 2018-12-03
Use replication factor in ReplicatedTableProvider

emolsson:
Fix issue #12

### DIFF
--- a/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/utils/ReplicatedTableProviderImpl.java
+++ b/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/utils/ReplicatedTableProviderImpl.java
@@ -32,6 +32,8 @@ public class ReplicatedTableProviderImpl implements ReplicatedTableProvider
     private static final String SIMPLE_STRATEGY = "org.apache.cassandra.locator.SimpleStrategy";
     private static final String NETWORK_TOPOLOGY_STRATEGY = "org.apache.cassandra.locator.NetworkTopologyStrategy";
 
+    private static final String SIMPLE_STRATEGY_REPLICATION_FACTOR = "replication_factor";
+
     private static final String SYSTEM_AUTH_KEYSPACE = "system_auth";
 
     private final Host myLocalhost;
@@ -85,7 +87,7 @@ public class ReplicatedTableProviderImpl implements ReplicatedTableProvider
 
     private boolean validateSimpleStrategy(Map<String, String> replication)
     {
-        int replicationFactor = Integer.parseInt(replication.get("replication_factor"));
+        int replicationFactor = Integer.parseInt(replication.get(SIMPLE_STRATEGY_REPLICATION_FACTOR));
 
         return replicationFactor > 1;
     }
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/utils/ReplicatedTableProviderImpl.java
```diff
@@ -52,14 +61,66 @@ public boolean accept(String keyspace)
...
 
-        if (tokenRanges.isEmpty())
+        if (keyspaceMetadata != null)
         {
+            Map<String, String> replication = keyspaceMetadata.getReplication();
+            String replicationClass = replication.get(STRATEGY_CLASS);
+
+            switch(replicationClass)
+            {
+                case SIMPLE_STRATEGY:
```
eperott: Will you always get class name _including_ packages here?

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/utils/ReplicatedTableProviderImpl.java
```diff
@@ -52,14 +61,66 @@ public boolean accept(String keyspace)
...
+                    return validateSimpleStrategy(replication);
+                case NETWORK_TOPOLOGY_STRATEGY:
+                    return validateNetworkTopologyStrategy(keyspace, replication);
+                default:
+                    LOG.warn("Replication strategy of type {} is not supported", replicationClass);
```
eperott: Can I get rid of this warning message if I've created my own custom strategy? If not, should it be at info-level?

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/utils/ReplicatedTableProviderImpl.java
```diff
@@ -52,14 +61,66 @@ public boolean accept(String keyspace)
...
+                    break;
+            }
+        }
+
+        return false;
+    }
+
+    private boolean validateSimpleStrategy(Map<String, String> replication)
+    {
+        int replicationFactor = Integer.parseInt(replication.get("replication_factor"));
```
eperott: "replication_factor" should be declared as a constant just like STRATEGY_CLASS

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/utils/ReplicatedTableProviderImpl.java
emolsson: The class name stored in system_schema.keyspaces where it uses Class#getName() to retrieve the name. So I believe this should always be the case.

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/utils/ReplicatedTableProviderImpl.java
emolsson: Not as of now, if you don't provide your own ReplicatedTableProvider.. But I think it make sense to have this as warning still since no repair would be scheduled for this table with the default providers.

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/utils/ReplicatedTableProviderImpl.java
emolsson: +1

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/utils/ReplicatedTableProviderImpl.java
eperott: Fair enough

PR #18, opened 2018-12-05
Four node for integration tests

emolsson:


osgi-integration/pom.xml
```diff
@@ -208,6 +208,28 @@
                                         <image>
                                             <name>cassandra:3.0</name>
                                             <alias>cassandra3</alias>
+                                            <run>
+                                                <memory>${it.cassandra.memory}</memory>
+                                                <memorySwap>-1</memorySwap>
+                                                <env>
+                                                    <LOCAL_JMX>no</LOCAL_JMX>
+                                                    <CASSANDRA_SEEDS>$${docker.container.seed.ip}</CASSANDRA_SEEDS>
```
itskarlsson: Why is this not be on the first node as well?

emolsson: I can't remember the exact details, but I believe there was some issue to reference the IP of the docker container in that way.

PR #27, opened 2018-12-11
Configurable repair locking type

emolsson:
Add configuration for standalone and OSGi


ecchronos-binary/src/resources/ecChronos.cfg
```diff
@@ -60,6 +60,13 @@
 ## Possible values are "parallel"
 #repair.parallelism=parallel
 
+## Which locking type to use for repairs.
+## "vnode" will lock each node involved in the repair individually and increase the number of
+## parallel repairs that can run in a single data center.
+## "data_center" will lock each data center involved in the repair and only allow a single repair per data center.
+## Possible values are "vnode", "data_center" and "data_center_and_vnode".
```
itskarlsson: datacenter
Also, put in a few words about what datacenter_and_vnode does just to be clear

PR #33, opened 2019-01-16
Remove incremental repair

emolsson:


core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairGroup.java
```diff
@@ -124,35 +123,17 @@ public String toString()
...
-        else
-        {
-            Set<Host> replicas = myReplicaRepairGroup.getReplicas();
-
-            if (!replicas.isEmpty())
-            {
-                builder.withReplicas(replicas);
-            }
+            builder.withTokenRanges(Collections.singletonList(range))
+                    .withReplicas(myReplicaRepairGroup.getReplicas());
```
etedpet: Since the builder is reused within this for-loop and the withReplicas set the "same value" every time, I guess it can be moved up a few lines (outside the for-loop) - where the other "static" values are set to the builder...

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairTask.java
```diff
@@ -211,30 +209,27 @@ public String toString()
...
+
         if (myReplicas != null)
         {
-            StringBuilder sb = new StringBuilder();
+            StringBuilder replicasStringBuilder = new StringBuilder();
 
             for (Host host : myReplicas)
             {
-                sb.append(host.getBroadcastAddress().getHostAddress()).append(',');
+                replicasStringBuilder.append(host.getBroadcastAddress().getHostAddress()).append(',');
```
etedpet: Maybe it does not matter that it is a "," after the last address in the list?
But may still be nice to use a stream/collect/joining to create the string of addresses... Something like this:
String hostAddresses = myReplicas.stream()
    .map(host -> host.getBroadcastAddress().getHostAddress())
    .collect(joining(","));

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairGroup.java
emolsson: Good catch, I'll fix that.

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairTask.java
emolsson: Looking in the Cassandra code it's parsed with a [StringTokenizer](https://github.com/apache/cassandra/blob/cassandra-3.0/src/java/org/apache/cassandra/repair/messages/RepairOption.java#L189). The StringTokenizer handles the last ",".

But I think your suggestion is correct so I'll change the implementation accordingly.

PR #34, opened 2019-01-21
Improve documentation

emolsson:


docs/SETUP.md
```diff
@@ -0,0 +1,74 @@
+# Setup
+
+## Preparation
+
+In order to allow ecChronos to run there are a few tables that needs to be present.
```
itskarlsson: This sounds like creation isn't automatic. Try: "In order to allow ecChronos to run there are a few tables that are created."

docs/SETUP.md
```diff
@@ -0,0 +1,74 @@
...
+The keyspace name is configurable and is `ecchronos` by default.
```
itskarlsson: The keyspace created is 'ecchronos' but can be overwritten with configuration options (Link to options)

docs/SETUP.md
```diff
@@ -0,0 +1,74 @@
...
+It is important that the keyspace is configured to replicate to all data centers.
+It is also highly recommended to use `NetworkTopologyStategy`.
+
+The required tables are shown below:
```
itskarlsson: created

docs/SETUP.md
itskarlsson: After some offline discussions I figured out that the creation is not automatic. I believe we should make this automatic. I understand that we need to face the UUID problem of creating tables concurrently, but we could simply include a script or something that can be run afterwards as a compromise.

Either way, this should not be solved in this patchset.

PR #46, opened 2019-03-14
Add State to each ScheduledJob

itskarlsson:
* Refactor runnable into getState
* Added finished state that marks jobs for deschedule


core/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/scheduling/DummyJob.java
```diff
@@ -22,7 +22,7 @@
 
 public class DummyJob extends ScheduledJob
 {
-    private volatile boolean hasRun = false;
+    volatile boolean hasRun = false;
```
emolsson: Nit: protected?

Feel free to leave like this as well.

itskarlsson: Let's stay with package private. More restrictive.

PR #49, opened 2019-03-14
Add RepairJobView

emolsson:


core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairJobView.java
```diff
@@ -0,0 +1,39 @@
...
+import com.ericsson.bss.cassandra.ecchronos.core.repair.state.RepairStateSnapshot;
+import com.ericsson.bss.cassandra.ecchronos.core.utils.TableReference;
+
+/**
+ * Read only view of a scheduled repair job.
+ */
+public interface RepairJobView
+{
+    /**
+     * Return the table this job is scheduled for.
```
etedpet: I think it is better to use the @return tag, than writing "Return" in the description... For "simple" methods like this maybe no special "description" is needed. 
The @return tag is a mandatory tag for all non-void methods (https://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#@return)

So what I'm trying to say is just change "Return" to "@return" :-)

(same for the other methods)

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairScheduler.java
```diff
@@ -37,4 +39,9 @@
      * @param tableReference The table to remove configuration for.
      */
     void removeConfiguration(TableReference tableReference);
+
+    /**
+     * Get a list of the currently scheduled (read-only) repair jobs.
```
etedpet: @return ?

also - what does "(read-only)" mean here? Is it needed?

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/TableRepairJob.java
```diff
@@ -61,16 +61,24 @@
...
     }
 
+    @Override
     public RepairConfiguration getRepairConfiguration()
     {
         return myRepairConfiguration;
     }
 
+    @Override
+    public RepairStateSnapshot getRepairStateSnapshot()
```
etedpet: Is there any test cases for this new method? (or any of the methods in the RepairJobView interface)?

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairSchedulerImpl.java
```diff
@@ -104,6 +106,15 @@ public void removeConfiguration(TableReference tableReference)
         myExecutor.execute(() -> handleTableConfigurationRemoved(tableReference));
     }
 
+    @Override
+    public List<RepairJobView> getCurrentRepairJobs()
```
etedpet: Test case for this method?

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairScheduler.java
emolsson: Removing read-only, this returns RepairJobView(immutable view) rather than the actual TableRepairJobs(can perform actions) which was the "read-only" part mentioned.

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairJobView.java
emolsson: Changing to "@return", I checked a bit on this and in the google coding guidelines this seemed to be preferred. But following oracle might be better.

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/TableRepairJob.java
emolsson: Added a test case to check all arguments of RepairJobView.

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairSchedulerImpl.java
emolsson: Added checks to existing test cases

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairJobView.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package com.ericsson.bss.cassandra.ecchronos.core.repair;
+
+import com.ericsson.bss.cassandra.ecchronos.core.repair.state.RepairStateSnapshot;
+import com.ericsson.bss.cassandra.ecchronos.core.utils.TableReference;
+
+/**
+ * Read only view of a scheduled repair job.
+ */
+public interface RepairJobView
```
itskarlsson: Food for thought: What benefits do you see having a interface define what variables will be output instead of adding a method into TableRepairJob that returns an instance of a view of itself? I feel that the latter would be a little more understandable for the price of some copying. May be worth investigating how it looks.

emolsson: Good suggestion, I've updated this.

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/TableRepairJob.java
```diff
@@ -22,12 +22,12 @@
 
 import com.ericsson.bss.cassandra.ecchronos.core.JmxProxyFactory;
 import com.ericsson.bss.cassandra.ecchronos.core.repair.state.ReplicaRepairGroup;
+import com.ericsson.bss.cassandra.ecchronos.core.scheduling.ScheduledJob;
 import com.ericsson.bss.cassandra.ecchronos.core.scheduling.ScheduledTask;
 import com.ericsson.bss.cassandra.ecchronos.core.repair.state.RepairState;
 import com.ericsson.bss.cassandra.ecchronos.core.repair.state.RepairStateSnapshot;
 import com.ericsson.bss.cassandra.ecchronos.core.utils.TableReference;
 import com.ericsson.bss.cassandra.ecchronos.core.metrics.TableRepairMetrics;
-import com.ericsson.bss.cassandra.ecchronos.core.scheduling.ScheduledJob;
```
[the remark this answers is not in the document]
emolsson: Done

PR #51, opened 2019-03-19
Add changelog

emolsson:


CHANGES.md
```diff
@@ -0,0 +1,13 @@
+# Changes
+
+## Version 2.0
```
eperott: Are we already moving towards a new major version? I.e, are there any non-bwc changes in the list. I suppose the filename changes qualify...?

With semantic versioning, shouldn't there be 3 positions in the version number?

CHANGES.md
```diff
@@ -0,0 +1,13 @@
...
+
+* Dynamic license header check - #37
```
eperott: Does it make sense to include "everything" here such as these things, or should we keep it down to "user facing" changes?

CHANGES.md
eperott: pom files indicate we're moving towards version 1.0.1

emolsson: Changed pom versions to 2.0.0 & updated the changlog with 2.0.0

CHANGES.md
emolsson: I think user/dev facing changes should be mentioned. I.e. not updating karaf test version, but if we change an interface.

eperott: Sounds reasonable.

Another way to slice would be to keep user facing stuff only in the changelog, and let developers use git log. No strong opinion though.

PR #66, opened 2019-04-30
Fix statistics when no table/data to repair

### LINKED ISSUE #59: Statistics writes NaN
DataRepairState.csv is generating NaN, example:

```
t,value
1556094425,NaN
1556094522,NaN
1556094582,NaN
1556094642,NaN
1556094651,NaN
```

This is likely happening when total node data size is 0. A similar issue could potentially happen for TableRepairState.csv as well if we don't have any tables to repair.

etedpet:
Fixes #59

### DIFF
--- a/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/metrics/NodeMetricHolder.java
+++ b/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/metrics/NodeMetricHolder.java
@@ -56,16 +56,12 @@ public class NodeMetricHolder implements Closeable
             @Override
             protected Ratio getRatio()
             {
-                int tables = 0;
-                double repairRatio = 0;
+                double averageRatio = myTableRepairRatio.values().stream()
+                        .mapToDouble(d -> d)
+                        .average()
+                        .orElse(1); // 100% when no tables to repair
 
-                for (Double ratio : myTableRepairRatio.values())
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/metrics/NodeMetricHolder.java
```diff
@@ -79,6 +85,11 @@ protected Ratio getRatio()
                         .map(Map.Entry::getKey)
                         .collect(Collectors.toList());
 
+                if (repairedTables.isEmpty())
```
etedpet: The hole method could be replaced with a single stream:

protected Ratio getRatio()
{
   return myTableRepairRatio.entrySet().stream()
            .filter(e -> e.getValue() > 0)
            .map(Map.Entry::getKey)
            .map(tableReference -> tableStorageStates.getDataSize(tableReference) * myTableRepairRatio.get(tableReference))
            .reduce(Double::sum)
            .map(size -> Ratio.of(size, tableStorageStates.getDataSize()))
            .orElse(RATIO_ONE);
}

But maybe that is "too much" ;-/
The NaN is solved at the end when the Optional sum is mapped into a Ratio OR the default value 1...

etedpet:
Will close this pull request and start another one against the 1.0 branch...

PR #69, opened 2019-05-06
Improve CONTRIBUTING.md

emolsson:


CONTRIBUTING.md
```diff
@@ -4,6 +4,25 @@ You are most welcome to create pull requests and issues to ecChronos.
 Before creating pull requests it is recommended to first discuss your idea with at least one of the owners of the repository.
 For keeping track of the history it is recommended that most communication is performed or summarized within pull requests or issues.
 
+## Branches
+
+Target the minimum possible branch version when fixing an issue.
```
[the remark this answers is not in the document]
emolsson: Hmm.. I'm thinking that if it is a bug fix then we should target ecchronos-1.0 when applicable. But if it's a new feature then we should target the master branch, etc.

Maybe a better way to phrase it would be:

> Target the lowest applicable version of ecChronos when fixing an issue.
> Bug fixes should be targeted to the lowest maintained version where the bug resides.
> New features should, in general, be added to the master branch.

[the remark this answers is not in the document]
etedpet: I think that would explain things better :-)

emolsson:
> The first sections of the project README.md has some slight overlap with the contents of this patch. Would it make sense to move all contributor related stuff into this file, and make the README more compelling to potential users (rather than contributors).

I'm not sure I see the overlap but are you thinking about i.e. running the test cases and similar? I think that could make sense to put in the CONTRIBUTING.md file instead.

eperott:
> I'm not sure I see the overlap but are you thinking about i.e. running the test cases and similar? I think that could make sense to put in the CONTRIBUTING.md file instead.

Yes, I was thinking about the "running tests" section. But IMO also things related to Eclipse setup, structure, architecture and build tools could be moved into the contributing doc. Then again, I'm just a guy with an opinion. :-)

README.md
```diff
@@ -12,65 +12,14 @@ See the [STRUCTURE.md](docs/STRUCTURE.md) and [ARCHITECTURE.md](docs/ARCHITECTUR
...
-
-## Deployment
-
-For deployment on a running system read [SETUP.md](docs/SETUP.md) for details on how to setup your system to enable ecChronos to run.
-
-## Built with
-
-* [Maven](https://maven.apache.org) - Dependency and build management
-* [docker-maven-plugin](https://github.com/fabric8io/docker-maven-plugin) - For integration tests
+For installation on a running system read [SETUP.md](docs/SETUP.md) for details on how to setup your system to enable ecChronos to run.
```
eperott: Rephrase?

CONTRIBUTING.md
```diff
@@ -4,46 +4,97 @@ You are most welcome to create pull requests and issues to ecChronos.
...
-
+### Builds
+    
 If you're looking to contribute to ecChronos it is recommended that you sign up on [Travis](https://travis-ci.org/) and [Coveralls](https://coveralls.io/) and enable builds and reports on your own fork of ecChronos.
 The builds required to merge a pull requests are contained within the [Travis configuration](.travis.yml) and include tests, code coverage as well as PMD checks.
 
+All checks needs to pass before merging a pull request.
+Sometimes the PMD rules does not match the goal of the project and in those cases it might be relevant to discuss removing or ignoring specific rules/violations.
+Currently there is a base of PMD rules used and that will evolve with the project,
+so if a rule seems to be detrimental rather helping the project,
```
eperott: ...rather than helping...?

README.md
emolsson: Done

CONTRIBUTING.md
emolsson: I think it's technically correct (the best kind) as the PMD rules are there to help the project. But I see your point so I rephrased this whole segment. :)

eperott: I was mostly referring to the missing "than", but this work too. :-)

Issue #70, opened 2019-05-16
Locking failures trigger to often


emolsson:
If we have ten tables that needs to be repaired within the same keyspace and the first one fails to lock a resource, instead of trying it with all the other nine tables we should cache the result and fail-fast.

emolsson:
Closed with PR #71

PR #71, opened 2019-05-20
Cache lock failure

emolsson:
Cache locking failures to avoid retrying the same lock over and over.

- Add lock cache in CASLockFactory.
- Add possibility for repair to fail-fast if locks are already known to be unavailable.

Fixes Issue #70

### DIFF
--- a/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/LockCache.java
+++ b/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/LockCache.java
@@ -48,6 +48,11 @@ public class LockCache
                 .build();
     }
 
+    public Optional<LockException> getCachedFailure(String dataCenter, String resource)
+    {
+        return getCachedFailure(new LockKey(dataCenter, resource));
+    }
+
     public DistributedLock getLock(String dataCenter, String resource, int priority, Map<String, String> metadata) throws LockException
     {
         LockKey lockKey = new LockKey(dataCenter, resource);
--- a/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/LockCache.java
+++ b/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/LockCache.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Telefonaktiebolaget LM Ericsson
+ * Copyright 2019 Telefonaktiebolaget LM Ericsson
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,12 +16,14 @@ package com.ericsson.bss.cassandra.ecchronos.core;
 
 import com.ericsson.bss.cassandra.ecchronos.core.exceptions.LockException;
 import com.ericsson.bss.cassandra.ecchronos.core.scheduling.LockFactory.DistributedLock;
+import com.google.common.base.Preconditions;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
@@ -95,12 +97,13 @@ public class LockCache
     static final class LockKey
     {
         private final String myDataCenter;
-        private final String myResource;
+        private final String myResourceName;
 
         LockKey(String dataCenter, String resource)
         {
+            Preconditions.checkNotNull(resource);
             myDataCenter = dataCenter;
-            myResource = resource;
+            myResourceName = resource;
         }
 
         @Override
@@ -108,19 +111,15 @@ public class LockCache
         {
             if (this == o) return true;
             if (o == null || getClass() != o.getClass()) return false;
-
             LockKey lockKey = (LockKey) o;
-
-            if (myDataCenter != null ? !myDataCenter.equals(lockKey.myDataCenter) : lockKey.myDataCenter != null) return false;
-            return myResource.equals(lockKey.myResource);
+            return Objects.equals(myDataCenter, lockKey.myDataCenter) &&
+                    myResourceName.equals(lockKey.myResourceName);
         }
 
         @Override
         public int hashCode()
         {
-            int result = myDataCenter != null ? myDataCenter.hashCode() : 0;
-            result = 31 * result + myResource.hashCode();
-            return result;
+            return Objects.hash(myDataCenter, myResourceName);
         }
     }
 }
--- a/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/LockCache.java
+++ b/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/LockCache.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2018 Telefonaktiebolaget LM Ericsson
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.ericsson.bss.cassandra.ecchronos.core;
+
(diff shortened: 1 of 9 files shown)


emolsson:
With a setup of 11 nodes (dc1:5, dc2:5, dc3:1) and three keyspaces:
ks1 = RF(dc1:3, dc3:1)
ks2 = RF(dc2:3, dc3:1)
ks3 = RF(dc1:3, dc2:3, dc3:1)

(and nine empty tables in each keyspace)

I ran everything using CCM and (11) local ecChronos instances. The reason to use rf=1 on one DC was to make sure that there was a lot of contention on that lock.

General CPU usage decreased from **~79.3%** to **~25,5%** with this patch applied.
Repair session latency (max) improved from **~2400ms** to **~200ms**.
Repair session latency (average) improved from **~167ms** to **46ms**
Repair session rate was similar from **~1.31** to **~1.43** sessions/s through the cluster.

I did some additional tests with other changes as well which increased the repair session rate by _x2.8_ for the empty tables. I'll create a new issue and PR for this (not stopping repair after the first session is completed). This didn't seem to have a large impact on CPU usage, at least when there is no data in the tables. I'll have to run some additional tests with actual data in the tables for that though.

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/LockCache.java
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright 2018 Telefonaktiebolaget LM Ericsson
```
emolsson: 2019

core/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/TestLockCache.java
```diff
@@ -0,0 +1,142 @@
+/*
+ * Copyright 2018 Telefonaktiebolaget LM Ericsson
```
emolsson: 2019

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/LockCache.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private void throwCachedLockException(LockException e) throws LockException
+    {
+        LOG.debug("Encountered cached locking failure, throwing exception", e);
+        throw e;
+    }
+
+    private Optional<LockException> getCachedFailure(LockKey lockKey)
+    {
+        return Optional.ofNullable(myFailureCache.getIfPresent(lockKey));
```
etedpet: I really like optionals, but maybe there is no meaning in doing it here...?
What you would want todo in getLock() is:
  cachedFailure.ifPresent(this::throwCachedLockException);
But that is not possible since the method throws a checked exception...

So here we wrap the value (cachedFailure) in a Optional, an then just use the isPresent() and get() methods on it... A simple cachedFailure != null would have "looked the same", but without this method...

I still like Optionals, but just pointing out that extra wrapping maybe is not needed here.

core/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/TestLockCache.java
```diff
@@ -0,0 +1,142 @@
...
+    {
+        assertThatExceptionOfType(LockException.class).isThrownBy(() -> myLockCache.getLock(DATA_CENTER, resource, PRIORITY, METADATA));
+    }
+
+    private void doReturnLockOnSupply() throws LockException
+    {
+        doReturnLockOnSupply(RESOURCE);
+    }
+
+    private void doReturnLockOnSupply(String resouce) throws LockException
```
etedpet: spelling of resource

core/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/TestLockCache.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    private void assertCacheThrowsException()
+    {
+        assertCacheThrowsException(RESOURCE);
+    }
+
+    private void assertCacheThrowsException(String resource)
+    {
+        assertThatExceptionOfType(LockException.class).isThrownBy(() -> myLockCache.getLock(DATA_CENTER, resource, PRIORITY, METADATA));
```
etedpet: Maybe we should test more than the class of the exception thrown? Don't we want the cache to throw the SAME exception as was thrown the first time?

core/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/TestLockCache.java
```diff
@@ -0,0 +1,142 @@
...
+    @Before
+    public void setup()
+    {
+        myLockCache = new LockCache(mockedLockSupplier);
+    }
+
+    @Test
+    public void testGetLock() throws LockException
+    {
+        doReturnLockOnSupply();
```
itskarlsson: I would find 'doReturnLockonGetLock()' to be a little clearer

core/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/TestLockCache.java
```diff
@@ -0,0 +1,142 @@
...
+        doReturnLockOnSupply(otherResource);
+
+        assertCacheThrowsException(RESOURCE);
+        assertExpectedLockIsRetrieved(otherResource);
+    }
+
+    @Test
+    public void testGetLockAfterCachedExceptionHasExpired() throws LockException, InterruptedException
+    {
+        myLockCache = new LockCache(mockedLockSupplier, 20, TimeUnit.MILLISECONDS);
```
itskarlsson: 20 could be reduced a bit for sake of test time.

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/LockCache.java
```diff
@@ -95,32 +97,29 @@ private void throwCachedLockException(LockException e) throws LockException
...
         {
             if (this == o) return true;
             if (o == null || getClass() != o.getClass()) return false;
-
             LockKey lockKey = (LockKey) o;
-
-            if (myDataCenter != null ? !myDataCenter.equals(lockKey.myDataCenter) : lockKey.myDataCenter != null) return false;
-            return myResource.equals(lockKey.myResource);
+            return Objects.equals(myDataCenter, lockKey.myDataCenter) &&
+                    myResourceName.equals(lockKey.myResourceName);
```
etedpet: Personally I like the "symmetry" by using Objects.equals on both members... Even though you need 1 extra method call on the stack when using it...
return Objects.equals(myDataCenter, that.myDataCenter) &&
                Objects.equals(myResourceName, that.myResourceName);

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/LockCache.java
```diff
@@ -95,32 +97,29 @@ private void throwCachedLockException(LockException e) throws LockException
...
         private final String myDataCenter;
-        private final String myResource;
+        private final String myResourceName;
 
         LockKey(String dataCenter, String resource)
         {
+            Preconditions.checkNotNull(resource);
             myDataCenter = dataCenter;
-            myResource = resource;
+            myResourceName = resource;
```
etedpet: checkNotNull also returns the value it checks... So you could do it on the same line as the assignment (with static import it gets kinda nice ;)
myResourceName = checkNotNull(resource);

emolsson: Thanks for the tip :)

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/LockCache.java
emolsson: Changed both implementations to use Objects.equals instead

core/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/repair/TestRepairLockFactoryImpl.java
```diff
@@ -165,23 +177,76 @@ public void testMultipleLocksOneFailing() throws LockException
                 .isThrownBy(() -> repairLockFactory.getLock(mockLockFactory, Sets.newHashSet(repairResourceDc1, repairResourceDc2), metadata, priority));
     }
 
+    @Test
+    public void testMultipleLocksOneHasCachedFailure() throws LockException
+    {
+        RepairResource repairResourceDc1 = new RepairResource("DC1", "my-resource-dc1");
+        RepairResource repairResourceDc2 = new RepairResource("DC2", "my-resource-dc2");
+        RepairLockFactoryImpl repairLockFactory = new RepairLockFactoryImpl();
+        Map<String, String> metadata = new HashMap<>();
```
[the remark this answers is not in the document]
emolsson: Done

PR #79, opened 2019-09-18
Reduce error levels to warn

emolsson:
These log messages are not critical and should be recoverable,
no need to log them on error level.

### DIFF
--- a/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/CASLockFactory.java
+++ b/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/CASLockFactory.java
@@ -222,7 +222,7 @@ public class CASLockFactory implements LockFactory, Closeable
         }
         catch (Exception e)
         {
-            LOG.error("Unable to retrieve metadata for resource {}", resource, e);
+            LOG.warn("Unable to retrieve metadata for resource {}", resource, e);
         }
 
         return null;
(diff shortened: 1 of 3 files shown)


eperott:
What's a good policy for choosing between WARN and ERROR? How about:
- WARN -> something failed but may work on next try - or we are close to reaching a limit of some sort.
- ERROR -> unrecoverable failure in some critical function - attention required

When considering the changes in this patch set, will ecChronos be able to recover in all those cases without a restart or change of settings? If "yes", then I agree this is a motivated change. If "no", then perhaps it makes more sense to keep things at ERROR level...

emolsson:
> What's a good policy for choosing between WARN and ERROR? How about:
> 
>     * WARN -> something failed but may work on next try - or we are close to reaching a limit of some sort.
> 
>     * ERROR -> unrecoverable failure in some critical function - attention required
> 
> 
> When considering the changes in this patch set, will ecChronos be able to recover in all those cases without a restart or change of settings? If "yes", then I agree this is a motivated change. If "no", then perhaps it makes more sense to keep things at ERROR level...

Release locks:
Everything should be recoverable, the locks use a TTL of 10 minutes, so if releasing it fails (DELETE IF) then it will be removed automatically later on.

Taking lock:
This will be retried later on, so this is only a problem if a quorum of the replica nodes are unavailable for a long time.

Issue #86, opened 2019-11-25
Slow query of repair_history at start-up


etedpet:
ecChronos queries system_distributed.repair_history to be able to schedule new repairs at starts-up. If the history table is "large" and contains a lot of tombstones a warning is printed in the cassandra logs:
WARN  o.a.c.d.ReadCommand$1MetricRecording:569 onClose Read 5000 live rows and 6144 tombstone cells for query SELECT * FROM system_distributed.repair_history WHERE keyspace_name, columnfamily_name = <ks>, <table> AND id <= 8be5a6ef-0790-11ea-7f7f-7f7f7f7f7f7f LIMIT 5000 (see tombstone_warn_threshold)

Also this query can be very slow since it iterates through all data with using the prepared statement:
"SELECT id, range_begin, range_end, status, participants FROM %s.%s WHERE keyspace_name=? AND columnfamily_name=? AND id <= maxTimeuuid(?)"

If the "last repaired at"-date is known (after startup), then the iteration is then limited to that date, using the following prepared statement:
"SELECT id, range_begin, range_end, status, participants FROM %s.%s WHERE keyspace_name=? AND columnfamily_name=? AND id >= minTimeuuid(?) and id <= maxTimeuuid(?)"

We should introduce a configurable limit for how long back in time this iteration should continue - even if "last repaired at" is unknown.

PR #87, opened 2019-11-26
Fix slow repair_history query (#86)

etedpet:


core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/state/RepairHistoryProviderImpl.java
```diff
@@ -49,28 +51,34 @@
...
+        myClock = clock;
     }
 
     @Override
     public Iterator<RepairEntry> iterate(TableReference tableReference, long to, Predicate<RepairEntry> predicate)
     {
-        ResultSet resultSet = execute(myRepairHistoryStatement.bind(tableReference.getKeyspace(), tableReference.getTable(), new Date(to)));
-
-        return RepairEntryIterator.create(resultSet.iterator(), predicate);
+        long from = myClock.millis() - myLookbackTime;
```
etedpet: This could potentially be a (theoretical) problem... IF myLookbackTime is a huge number (e.g. Long.MAX_VALUE), then java.util.Date(from) will be in the future (instead of being far in the past)... Maybe that is not a problem in practice?
But if we should handle this - what is the from-value we should limit to?
0? - in that case we cannot find entries being created before epoch (1970), but that is maybe OK...

emolsson: I'd be ok with introducing a check for it (in the next method), something like:

```
if (!startDate.before(endDate))
{
  throw new IllegalArgumentException("");
}
```

But then we should add a test case for it as well.

---

Also I think this could be an issue if "to" is sent in lower than "from".

etedpet: added check

PR #92, opened 2019-12-16
Introduce equalsverifier tests

etedpet:
To verify that implemented equals and hashCode methods
follow the "equals contract".


etedpet:
Hmmm. There is some issue with Travis and the unit tests...
The tests works fine locally and when pushing to my private clone (origin) the tests passed.
But when a pull request was created the unit tests fail. Has to be investigated...

PR #102, opened 2020-03-23
Refactor deleteTableSchedule (#101)

itskarlsson:
Break out delete logic to avoid using ConcurrentHashMap

Fixes issue #101

### DIFF
--- a/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairSchedulerImpl.java
+++ b/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairSchedulerImpl.java
@@ -15,6 +15,7 @@
 package com.ericsson.bss.cassandra.ecchronos.core.repair;
 
 import java.io.Closeable;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairSchedulerImpl.java
```diff
@@ -15,6 +15,7 @@
 package com.ericsson.bss.cassandra.ecchronos.core.repair;
 
 import java.io.Closeable;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
```
emolsson: Remove obsolete dependency?

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairSchedulerImpl.java
```diff
@@ -127,7 +129,8 @@ private void createTableSchedule(TableReference tableReference, RepairConfigurat
 
         if (oldTableRepairJob != null)
         {
-            deleteTableSchedule(tableReference);
+            ScheduledJob job = myScheduledJobs.remove(tableReference);
```
emolsson: I think it would be better if the code is changed to:

```
TableRepairJob oldTableRepairJob = myScheduledJobs.remove(tableReference);

if (oldTableRepairJob != null)
{
  descheduleTableJob(oldTableRepairJob);
}
```

itskarlsson: Good catch!

itskarlsson: Just realized, we can remove the null check too

Issue #103, opened 2020-04-03
Add a code style to project


itskarlsson:
It would be good to decide on a code style and enforce it.

PR #105, opened 2020-04-03
Add a code style to project (#103)

itskarlsson:
Fixes issue #103

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,10 +19,10 @@ New features should, in general, be added to the master branch.
 
 ### Code Style
 
-Codestyle TBD.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -19,10 +19,10 @@ New features should, in general, be added to the master branch.
 
 ### Code Style
 
-Codestyle TBD.
+This project uses the cassandra code style which is based on Sun’s Java coding conventions. Formatting rules for eclipse can be found [here](https://github.com/Ericsson/ecchronos/blob/ecchronos-1.0/code_style.xml).
```
eperott, line 22: Once sentence per line?

I don't think the link works like that. You may have to use rawgit. Checkout this example that I did for the performance charts in ecAudit. https://github.com/Ericsson/ecaudit/blob/release/c2.2/doc/description.md#performance

Or pehaps it is fine to just say that the file can be found at the root of the repo?

itskarlsson: I'll try to add a relative path and see if that works.

PR #112, opened 2020-05-13
Range split

### LINKED ISSUE #96: Split/merge token range history
The current mechanism for both keeping track of and running repairs is based on the token ranges of the node. This is heavily relying on a large number of virtual nodes in order to split repairs.

An alternative to this would be to split and merge both repair history and token ranges as needed. Basically that the repair tasks are performed with smaller token ranges which are later merged together again when parsing the repair history. This would remove the need to use virtual nodes by having a configurable "token range size" instead.

Repair history for ranges that are adjacent could be merged i.e. if they are performed within an hour of each other or if both are outside of the repair interval. The merged result would retain the timestamp of the "oldest" entry.

Additionally the "token range size" could be dynamic - based on a "target repair size". 
`target token range = total token range / (current disk usage / target repair size)`
If a table has 10GB of data and we specify a target repair size of 1GB we would divide the total token range in 10 equal pieces while assuming that data is uniformly distributed across the token range.

emolsson:
Closes #96

### DIFF
--- a/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/RepairProperties.java
+++ b/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/RepairProperties.java
@@ -17,6 +17,7 @@ package com.ericsson.bss.cassandra.ecchronos.application;
 import com.ericsson.bss.cassandra.ecchronos.core.repair.RepairConfiguration;
 import com.ericsson.bss.cassandra.ecchronos.core.repair.RepairLockType;
 import com.ericsson.bss.cassandra.ecchronos.core.repair.RepairOptions;
+import com.ericsson.bss.cassandra.ecchronos.core.utils.UnitConverter;
 
 import java.util.Locale;
 import java.util.Properties;
@@ -31,6 +32,7 @@ public final class RepairProperties
     private static final String CONFIG_REPAIR_LOCK_TYPE = "repair.lock.type";
     private static final String CONFIG_REPAIR_UNWIND_RATIO = "repair.unwind.ratio";
     private static final String CONFIG_REPAIR_HISTORY_LOOKBACK_BASE = "repair.history.lookback";
+    private static final String CONFIG_REPAIR_TARGET_REPAIR_SIZE = "repair.size.target";
 
     private static final String DAYS = "days";
     private static final String DEFAULT_REPAIR_INTERVAL_TIMEUNIT = DAYS;
@@ -44,6 +46,7 @@ public final class RepairProperties
     private static final String DEFAULT_REPAIR_UNWIND_RATIO = Double.toString(RepairConfiguration.NO_UNWIND);
     private static final String DEFAULT_REPAIR_HISTORY_LOOKBACK_TIMEUNIT = DAYS;
     private static final String DEFAULT_REPAIR_HISTORY_LOOKBACK_DURATION = "30";
+    private static final String DEFAULT_REPAIR_TARGET_REPAIR_SIZE = "";
 
     private final long myRepairIntervalInMs;
     private final RepairOptions.RepairParallelism myRepairParallelism;
@@ -52,10 +55,12 @@ public final class RepairProperties
     private final RepairLockType myRepairLockType;
     private final double myRepairUnwindRatio;
     private final long myRepairHistoryLookbackInMs;
+    private final long myTargetRepairSizeInBytes;
 
     private RepairProperties(long repairIntervalInMs, RepairOptions.RepairParallelism repairParallelism,
                              long repairAlarmWarnInMs, long repairAlarmErrorInMs, RepairLockType repairLockType,
-                             double repairUnwindRatio, long repairHistoryLookbackInMs)
+                             double repairUnwindRatio, long repairHistoryLookbackInMs,
+                             long targetRepairSizeInBytes)
     {
         myRepairIntervalInMs = repairIntervalInMs;
         myRepairParallelism = repairParallelism;
@@ -64,6 +69,7 @@ public final class RepairProperties
         myRepairLockType = repairLockType;
         myRepairUnwindRatio = repairUnwindRatio;
         myRepairHistoryLookbackInMs = repairHistoryLookbackInMs;
+        myTargetRepairSizeInBytes = targetRepairSizeInBytes;
     }
 
     public long getRepairIntervalInMs()
@@ -99,6 +105,11 @@ public final class RepairProperties
     public long getRepairHistoryLookbackInMs()
     {
         return myRepairHistoryLookbackInMs;
+    }
+
+    public long getTargetRepairSizeInBytes()
+    {
+        return myTargetRepairSizeInBytes;
     }
 
     @Override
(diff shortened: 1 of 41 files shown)


application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/RepairProperties.java
```diff
@@ -145,8 +156,24 @@ public static RepairProperties from(Properties properties) throws ConfigurationE
         long repairHistoryLookbackInMs = parseTimeUnitToMs(properties, CONFIG_REPAIR_HISTORY_LOOKBACK_BASE,
                 DEFAULT_REPAIR_HISTORY_LOOKBACK_DURATION, DEFAULT_REPAIR_HISTORY_LOOKBACK_TIMEUNIT);
 
+        long targetRepairSizeInBytes = Long.MAX_VALUE;
```
itskarlsson, line 159: Perhaps add 
`private static final long DEFAULT_TARGET_REPAIR_SIZE_IN_BYTES = Long.MAX_VALUE;`
like you have done in the test class and use it here.

core.osgi/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/osgi/DefaultRepairConfigurationProviderComponent.java
```diff
@@ -61,13 +62,19 @@ public synchronized void activate(Configuration configuration)
         if (configuration.enabled())
         {
             long repairInterval = configuration.repairIntervalSeconds();
+            long repairSizeInBytes = Long.MAX_VALUE;
```
itskarlsson, line 65: Nit: `private static final long DEFAULT_TARGET_REPAIR_SIZE_IN_BYTES = Long.MAX_VALUE;`

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/exceptions/InternalException.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package com.ericsson.bss.cassandra.ecchronos.core.exceptions;
+
+/**
+ * An exception indicating an internal issue has occured.
+ * This typically indicates an unexpected bug in the software.
+ */
+public class InternalException extends RuntimeException
+{
+    public InternalException(String message)
```
itskarlsson, line 23: Add a serialVersionUID

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairConfiguration.java
```diff
@@ -23,12 +23,14 @@
 public class RepairConfiguration
 {
     public static final double NO_UNWIND = 0.0d;
+    public static final long FULL_REPAIR_SIZE = Long.MAX_VALUE;
```
itskarlsson, line 26: Shouldn't it be assigned to DEFAULT_TARGET_REPAIR_SIZE_IN_BYTES first and then to FULL_REPAIR SIZE?
`public static final long FULL_REPAIR_SIZE = DEFAULT_TARGET_REPAIR_SIZE_IN_BYTES;`

Also, do we need two fields for one being private and one being public? Can't it be the same field?

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairConfiguration.java
```diff
@@ -207,6 +218,22 @@ public Builder withRepairUnwindRatio(double repairUnwindRatio)
...
 
+        /**
+         * Set the target repair size in bytes.
+         *
+         * This is used to perform sub range repairs within virtual nodes.
+         * The sub ranges will be calculated based on how much data is in the table versus how large the target size
+         * per repair session is.
+         *
+         * @param targetRepairSizeInBytes The target data per repair session
+         * @return The builder
```
itskarlsson, line 229: Nit: Is the @return necessary on all of these builder functions?

docs/ARCHITECTURE.md
```diff
@@ -61,4 +61,47 @@ When a table is able to run repair the RepairState calculates the next tokens to
...
+(96, 98]  
+(98, 100]
+
+### Details
+
+Sub-ranges are handled in two parts, one part is while building an internal state of the repair history and the other is while performing the repairs.
+
+#### Repair history
+
+While building the internal repair history state all sub-ranges which are fully contained within a local virtual node is collected from the repair history.
```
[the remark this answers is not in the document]
emolsson: Done

docs/ARCHITECTURE.md
```diff
@@ -61,4 +61,47 @@ When a table is able to run repair the RepairState calculates the next tokens to
...
+The sub ranges would be:  
+(0, 2]  
+(2, 4]  
+...  
+(96, 98]  
+(98, 100]
+
+### Details
+
+Sub-ranges are handled in two parts, one part is while building an internal state of the repair history and the other is while performing the repairs.
```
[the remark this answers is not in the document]
emolsson: Done

docs/ARCHITECTURE.md
```diff
@@ -61,4 +61,47 @@ When a table is able to run repair the RepairState calculates the next tokens to
 The calculation is performed by the [VnodeRepairGroupFactory](../core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/state/VnodeRepairGroupFactory.java) by default.
 The TableRepairJob then generates [RepairGroups](../core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairGroup.java) which are snapshots from how the state was when it was calculated.
 When the RepairGroup is executed it will generate one [RepairTask](../core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairTask.java) per token range to repair.
-The RepairTask is the class that will perform the repair.
\ No newline at end of file
+The RepairTask is the class that will perform the repair.
+
+## Sub-range repairs
+
+As of #96 the repair scheduler in ecChronos has support for sub range repairs within virtual nodes.
```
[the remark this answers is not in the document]
emolsson: Done

docs/ARCHITECTURE.md
```diff
@@ -61,4 +61,47 @@ When a table is able to run repair the RepairState calculates the next tokens to
...
+This is activated by specifying a target repair size in the [RepairConfiguration](../core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairConfiguration.java).
+For the standalone version the option is called `repair.size.target` in the configuration file.
+Each sub-range repair session will aim to handle the target amount of data.
+
+*Note: Without this option specified the repair mechanism will handle full virtual nodes only (including how it interprets the repair history)*  
+*Note: The target repair size is assuming a uniform data distribution across partitions on the local node*
+
+### Example
```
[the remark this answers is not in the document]
emolsson: Yes and no.. The problem is that the algorithm kind of assumes that each token in the ring contains roughly the same amount of data. I don't think it's possible (without looking at the data) to decide how much each token actually contains.

The full mechanism is just an abstraction layer for the user where you specify a size-based "token range size" rather than specifying how many tokens you want to repair in each session. This also gives a nice benefit in that you can specify a standard amount of data you want to handle in each repair session so that small tables automatically gets repaired in fewer sessions while larger tables get split up more.

ecchronos-binary/src/resources/ecChronos.cfg
```diff
@@ -88,6 +88,16 @@
 #repair.history.lookback.time=30
 #repair.history.lookback.time.unit=days
 
+### Repair target size
+## Set a target for how much data each repair session should process.
+## This is an estimation assuming uniform data distribution among partition keys.
+## The value should be either a number or a number with a postfix:
```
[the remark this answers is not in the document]
emolsson: Done

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/RepairProperties.java
emolsson: Done

emolsson: Or actually, I referenced RepairConfiguration.FULL_REPAIR_SIZE instead to make the default purpose clearer.

core.osgi/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/osgi/DefaultRepairConfigurationProviderComponent.java
emolsson: Same as in RepairProperties, referencing RepairConfiguration.FULL_REPAIR_SIZE

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/exceptions/InternalException.java
emolsson: Done

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairConfiguration.java
emolsson: I would like to keep them separate as they are used for different things.
FULL_REPAIR_SIZE is a magic number used to specify "don't use the sub-range repair feature". The default value might change over time and should IMO then be a separate parameter. FULL_REPAIR_SIZE is used for both repair history and range split in order to activate the feature.

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairConfiguration.java
emolsson: When you run the javadoc target you get the following if you remove it:

```
[WARNING] ~/workspaces/ecchronos/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairConfiguration.java:229: warning: no @return
[WARNING] public Builder withTargetRepairSizeInBytes(long targetRepairSizeInBytes)
[WARNING] ^
```

PR #114, opened 2020-05-14
Support symlink creation of binary file

simonfontana:
Make it possible to create a symlink of the ecChronos binary file.
For example: ln -s /opt/ecchronos/bin/ecChronos /usr/bin/ecChronos

### DIFF
--- a/ecchronos-binary/src/bin/ecChronos.sh
+++ b/ecchronos-binary/src/bin/ecChronos.sh
@@ -15,8 +15,10 @@
 #
 
 if [ "x$ECCHRONOS_HOME" = "x" ]; then
-  ECCHRONOS_HOME="`dirname "$0"`/.."
+    ECCHRONOS_HOME=$(dirname $(readlink -f $0))/..
 fi
+
+cd $ECCHRONOS_HOME
 
 CLASSPATH="$ECCHRONOS_HOME"/conf/
 JVM_ENV=-Decchronos.config="$ECCHRONOS_HOME"/conf/ecChronos.cfg


emolsson:
Thanks for the PR, you're right that readlink should probably be used here. Could you open this PR towards `ecchronos-1.0` branch to make sure it works on all versions? I'll take care of merging it forward!

---

I got the following exception while trying to run it from a link:
```
a15:45:35.299 [main] ERROR c.e.b.c.e.application.ECChronos - Unable to start ecChronos
com.ericsson.bss.cassandra.ecchronos.application.ConfigurationException: Statistics directory './statistics' does not exist
	at com.ericsson.bss.cassandra.ecchronos.application.StatisticsProperties.from(StatisticsProperties.java:54)
```

I think this can be solved quite easily by adding:
```
JVM_ENV="$JVM_ENV -Decchronos.home=$ECCHRONOS_HOME"
```
to `ecc.sh` and then change the following in `StatisticsProperties.java`:


```
String homeDirectory = System.getProperty("ecchronos.home", "");
String defaultDirect = homeDirectory + DEFAULT_STATISTICS_DIRECTORY;

boolean enabled = Boolean.parseBoolean(properties.getProperty(CONFIG_STATISTICS_ENABLED, DEFAULT_STATISTICS_ENABLED));
File statisticsDirectory = new File(properties.getProperty(CONFIG_STATISTICS_DIRECTORY, defaultDirect));
```

Additionally it's probably good to update the default for normal/debug logs in logback, e.g:
```
<file>ecc.debug.log</file>
<fileNamePattern>ecc.debug.log.%d{yyyy-MM-dd}.%i.gz</fileNamePattern>
```
to:
```
<file>${ecchronos.home}ecc.debug.log</file>
<fileNamePattern>${ecchronos.home}ecc.debug.log.%d{yyyy-MM-dd}.%i.gz</fileNamePattern>
```
To avoid that it creates the log files where you are.

simonfontana:
> Could you open this PR towards ecchronos-1.0 branch to make sure it works on all versions? 

I wasn't sure if it was going to work since the file name had changed. But I just tried a forward merge with the patch on 1.0 and that works fine so I'll do that. 

> I got the following exception while trying to run it from a link

Got the same error when I moved out of the ecchronos directory. Will follow your suggestions and fix these issues in the same commit.

simonfontana:
So I decided to use a simpler, and in my opinion nicer solution, of simply cd into the ecChronos home directory. Hopefully you agree, otherwise I can change it to your proposed solution with editing StatisticsProperties and logback configuration.

PR #116, opened 2020-05-16
Add JVM options file

simonfontana:
Make it possible to set JVM options from file.

It's currently possible to set JVM options from command-line but I think this solution (which I copied from Cassandra project) is more elegant.

I only added a few example values which I know are good to set, if you know any other write them here and I'll add them.

### DIFF
--- a/ecchronos-binary/src/resources/jvm.options
+++ b/ecchronos-binary/src/resources/jvm.options
@@ -1,0 +1,36 @@
+#################################################################
+#                             jvm.options                       #
+#                                                               #
+# - all flags defined here will be used by to startup the JVM   #
+# - one flag should be specified per line                       #
+# - lines that do not start with '-' will be ignored            #
+# - only static flags are accepted (no variables or parameters) #
+#################################################################
+
+########################
+# GENERAL JVM SETTINGS #
+########################
+
+# Make sure all memory is faulted and zeroed on startup.
+# This helps prevent soft faults in containers and makes
+# transparent hugepage allocation more effective.
+#-XX:+AlwaysPreTouch
+
+# Sets the number of threads used during parallel phases of the garbage collectors.
+# The default value varies with the platform on which the JVM is running.
+# Having this value unset on JDK 8 and earlier versions may cause issues.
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -440,6 +440,7 @@ limitations under the License.
                             <exclude>**/*.txt</exclude>
                             <exclude>**/pom.xml.tag</exclude>
                             <exclude>**/pom.xml.releaseBackup</exclude>
+                            <exclude>**/*.options</exclude>
```
emolsson, line 443: Another option could be to add mapping like a few lines above:

```
<mapping>
  <options>SCRIPT_STYLE</options>
</mapping>
```

The current solution you have made is to add a header and then ignore the file from the check. I would prefer that we add the correct style for the file instead.

ecchronos-binary/src/resources/jvm.options
```diff
@@ -0,0 +1,45 @@
...
+# GENERAL JVM SETTINGS #
+########################
+
+# Make sure all memory is faulted and zeroed on startup.
+# This helps prevent soft faults in containers and makes
+# transparent hugepage allocation more effective.
+#-XX:+AlwaysPreTouch
+
+#################
+# HEAP SETTINGS #
```
emolsson, line 35: Another useful flag could be:

`-XX:ParallelGCThreads=1` to avoid situation like https://issues.apache.org/jira/browse/CASSANDRA-14475 / https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6858051

pom.xml
simonfontana: Right, I wasn't really sure what I was doing. Your solution proposal makes more sense. Fixed.

ecchronos-binary/src/resources/jvm.options
simonfontana: Done

pom.xml
simonfontana: As we agreed in private conversation I reverted to excluding .options files and removed the copyright header.

Issue #117, opened 2020-05-20
Repair policies are ignored within a task (RepairGroup)


emolsson:
Currently run policies are used to prevent a job from running. Since the current structure is
`Repair Job -> *Repair Group -> *Repair Task`
in case there is a large table being repaired this could prevent the run policy from taking effect potentially for hours.

Ideally specific targeted repair policies should prevent any new repair tasks from being executed. This would require restructuring policies to bind to the jobs rather than the schedule manager (potentially both).

A good first step for 1.x versions of ecc could be to at least prevent new repair groups from starting which should be relatively trivial.

emolsson:
After looking a bit closer I think it would be possible to fully achieve this in the 1.x branches by adding an optional list of repair policies to TableRepairJob directly. This should not break backwards compatibility and behave as before unless repair policies are added.

PR #118, opened 2020-05-20
Check run policy after tasks and sub-tasks

emolsson:
Add repair policy to TableRepairJob(RepairGroup) and check the policy
before running each RepairTask. Similar the policy will now be
consulted before running a ScheduledTask (RepairGroup) in order to
reduce the delay from policy activation to application.

Closes issue #117

### DIFF
--- a/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairGroup.java
+++ b/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairGroup.java
@@ -121,15 +121,7 @@ public class RepairGroup extends ScheduledTask
 
     private boolean shouldContinue()
     {
-        for (TableRepairPolicy repairPolicy : myRepairPolicies)
-        {
-            if (!repairPolicy.shouldRun(myTableReference))
-            {
-                return false;
-            }
-        }
-
-        return true;
+        return myRepairPolicies.stream().allMatch(repairPolicy -> repairPolicy.shouldRun(myTableReference));
     }
 
     @Override
--- a/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairGroup.java
+++ b/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairGroup.java
@@ -27,12 +27,7 @@ import com.google.common.annotations.VisibleForTesting;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
-import java.util.Collection;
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairGroup.java
```diff
@@ -101,6 +119,19 @@ public boolean execute()
         return successful;
     }
 
+    private boolean shouldContinue()
+    {
+        for (TableRepairPolicy repairPolicy : myRepairPolicies)
```
itskarlsson, line 124: Nit:
I believe this could be a little cleaner with streams. Something like this:
`myRepairPolicies.stream().anyMatch(policy::shouldRun(myTableReference))`

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairSchedulerImpl.java
```diff
@@ -227,6 +230,12 @@ public Builder withRepairLockType(RepairLockType repairLockType)
             return this;
         }
 
+        public Builder withRepairPolices(Collection<TableRepairPolicy> tableRepairPolicies)
```
itskarlsson, line 233: Policies

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/TableRepairPolicy.java
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.ericsson.bss.cassandra.ecchronos.core.repair;
+
+import com.ericsson.bss.cassandra.ecchronos.core.utils.TableReference;
+
+/**
+ * Interface for policies that can be used to control if repairs should continue to run.
```
itskarlsson, line 20: Nit:
 Interface for policies that can be used to control if repairs should **run.**

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairGroup.java
emolsson: +1, but I did it as:
```
myRepairPolicies.stream().allMatch(repairPolicy -> repairPolicy.shouldRun(myTableReference));
```
As we need to make sure it should run according to all policies.

PR #125, opened 2020-09-02
New PMD

emolsson:
* Use new PMD version
* Ensure project is compiled before pmd run

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ cache:
 matrix:
   include:
   - name: "Style check"
-    env: TEST_SUITE='com.mycila:license-maven-plugin:check pmd:pmd pmd:cpd pmd:check pmd:cpd-check javadoc:jar'
+    env: TEST_SUITE='compile com.mycila:license-maven-plugin:check pmd:pmd pmd:cpd pmd:check pmd:cpd-check javadoc:jar'
   - name: "Unit tests"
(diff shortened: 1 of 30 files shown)


itskarlsson (commented):
I'd be in favor of modifying the CloseResource rule so we do not have to add a bunch of NOPMD

emolsson:
I agree. Do you have a suggestion on how that can be done? I tried modifying the rule without success.

I believe this might also be an issue as we do not use try-with-resources or try-finally to close them (we create them at one place and close them somewhere else).

emolsson:
I agree. Do you have a suggestion on how that can be done? I tried modifying the rule without success.

I believe this might also be an issue as we do not use try-with-resources or try-finally to close them (we create them at one place and close them somewhere else).

itskarlsson:
There are some warnings too when building:
```
[WARNING] Unable to exclude rules [InvalidSlf4jMessageFormat] from ruleset reference category/java/errorprone.xml; perhaps the rule name is mispelled or the rule doesn't exist anymore?
[WARNING] Removed misconfigured rule: LoosePackageCoupling  cause: No packages or classes specified
```
Simply add to design
<exclude name="LoosePackageCoupling" />
and remove
<exclude name="InvalidSlf4jMessageFormat" />

itskarlsson (approved):

PR #135, opened 2020-09-22
Reduce OSGi test time

emolsson:
Use PerClass instead of PerMethod and wait until the state is
reset between test cases to avoid re-creating everything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ matrix:
   - name: "Unit tests"
     env: TEST_SUITE='test jacoco:report coveralls:report'
   - name: "OSGi integration"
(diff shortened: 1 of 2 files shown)


itskarlsson:
What is blocking us from running perSuite instead? Wouldn't that save even more time if it is possible?

emolsson:
> What is blocking us from running perSuite instead? Wouldn't that save even more time if it is possible?

I tried using the PerSuite option but two OSGi containers seemed to be created regardless. It looks like it might be necessary to create a wrapping test suite containing the two test classes and then changing the failsafe plugin configuration. As agreed offline the current solution (PerClass) is good enough for now.

Issue #130, opened 2020-09-11
Start application if Cassandra is in UJ


simonfontana:
Currently ecChronos initially tries to connect to Cassandra by CQL. If the connection fails ecChronos process will exit with error. 

EcChronos should recognize and wait if Cassandra is joining the cluster.

simonfontana:
If I understand correctly the implementation only adds a timeout for how long ecChronos should wait for CQL before throwing an error? I think it would be preferable if ecChronos could behave differently depending on if Cassandra is joining or not. When joining I would want a timeout of, for example, 2 hours, but when it's not joining I want 10 seconds. 

To make it clear, I'm okay with the current solution. Just wanted to give a suggestion on how and why I think we can still improve this feature. Maybe it's more difficult than I realize and it won't be worth it?

PR #143, opened 2020-10-15
Add configurable timeout for cql connections

itskarlsson:
Closes issue #130

### DIFF
--- a/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
+++ b/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
@@ -83,7 +83,8 @@ public class DefaultNativeConnectionProvider implements NativeConnectionProvider
                 }
                 catch (InterruptedException e1)
                 {
-                    throw new RuntimeException("Unexpected interrupt", e);
+                    // Should never occur
+                    throw new RuntimeException("Unexpected interrupt", e); //NOPMD
                 }
             }
         }
--- a/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
+++ b/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
@@ -19,10 +19,8 @@ import java.util.function.Supplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.datastax.driver.core.ExtendedAuthProvider;
-import com.datastax.driver.core.Host;
-import com.datastax.driver.core.SSLOptions;
-import com.datastax.driver.core.Session;
+import com.datastax.driver.core.*;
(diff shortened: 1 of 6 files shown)


application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
```diff
@@ -53,14 +51,45 @@ public DefaultNativeConnectionProvider(Config config, Supplier<Security.CqlSecur
...
 
-        myLocalNativeConnectionProvider = LocalNativeConnectionProvider.builder()
+        LocalNativeConnectionProvider.Builder nativeConnectionBuilder = LocalNativeConnectionProvider.builder()
                 .withLocalhost(host)
                 .withPort(port)
                 .withAuthProvider(authProvider)
-                .withSslOptions(sslOptions)
+                .withSslOptions(sslOptions);
+        establishTemporaryConnection(LocalNativeConnectionProvider.Builder.fromBuilder(nativeConnectionBuilder).build(),
+                host, port, nativeConfig.getTimeout());
```
emolsson, line 60: LocalNativeConnectionProvider.Builder.fromBuilder(...) seems a bit "into the internals".

Isn't it good enough to pass in the builder here and run "builder.build()" in the while loop?

That way we could also set myLocalNativeConnectionProvider to the result of the method immediately. The current implementation would leak a cluster/session object when finished.

connection.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/connection/impl/LocalNativeConnectionProvider.java
```diff
@@ -157,7 +157,7 @@ private static String resolveLocalDataCenter(Builder builder)
             throw new IllegalStateException("Unable to find local data center");
         }
 
-        private static Cluster.Builder fromBuilder(Builder builder)
+        public static Cluster.Builder fromBuilder(Builder builder)
```
emolsson, line 160: If the previously mentioned change is done this change shouldn't be necessary anymore.

emolsson (commented):
I'm not sure it's clear from the issue description but I believe a main part of the request to add this functionality was to make ecc respond to health-check requests while the Cassandra instance we are connecting to is bootstrapping.

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
```diff
@@ -56,39 +60,38 @@ public DefaultNativeConnectionProvider(Config config, Supplier<Security.CqlSecur
...
             try
             {
-                return cluster.connect();
+                return builder.build();
             }
             catch (NoHostAvailableException | IllegalStateException e)
             {
                 try
                 {
                     LOG.warn("Unable to connect through CQL using {}:{}, retrying", host, port);
```
emolsson, line 83: If we don't have any timeout time specified I don't think we should perform the sleep here.

emolsson (approved):

Issue #127, opened 2020-09-11
Expose metrics in REST API


simonfontana:
Expose metrics with Prometheus support through an endpoint in the REST API.

Metrics should be exposed and secured with TLS (mutual authentication).

itskarlsson:
Broke out the TLS part into a separate issue. (#147)

PR #146, opened 2020-10-21
Expose Metrics over REST in Prometheus format

itskarlsson:
Closes issue #127


core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/metrics/TableRepairMetricsImpl.java
```diff
@@ -131,13 +133,24 @@ public Builder withReportInterval(long reportInterval, TimeUnit timeUnit)
...
+        }
+
         public TableRepairMetricsImpl build()
         {
             if (myTableStorageStates == null)
             {
                 throw new IllegalArgumentException("Table storage states cannot be null");
             }
 
+            if (myMetricRegistry == null)
```
emolsson, line 149: Nit-picking:

Recently we have started using a pattern like the following in the constructors:
`
myMetricRegistry = Preconditions.checkNotNull(builder.myMetricRegistry, "Metric registry cannot be null");
`
Which IMO is cleaner than if and manual throws.

emolsson (approved):

emolsson (approved):

PR #160, opened 2020-11-04
Remove keyspace/table from id rest call

itskarlsson:


rest/src/main/java/com/ericsson/bss/cassandra/ecchronos/rest/RepairManagementRESTImpl.java
```diff
@@ -98,13 +98,12 @@ public String tableStatus(@PathVariable String keyspace, @PathVariable String ta
...
-    @GetMapping(ENDPOINT_PREFIX + "/status/keyspaces/{keyspace}/tables/{table}/ids/{id}")
-    public String jobStatus(@PathVariable String keyspace, @PathVariable String table, @PathVariable String id)
+    @GetMapping(ENDPOINT_PREFIX + "/status/ids/{id}")
+    public String jobStatus(@PathVariable String id)
     {
         try
         {
-            Optional<RepairJobView> repairJobView = getCompleteRepairJob(forTable(keyspace, table),
-                    UUID.fromString(id));
+            Optional<RepairJobView> repairJobView = getCompleteRepairJob(UUID.fromString(id));
```
emolsson, line 106: Nit:

Not really caused by this patch, but looking at it we should probably throw something like:
```
throw new ResponseStatusException(BAD_REQUEST, "Malformed job UUID", e);
```

in the try/catch below.

rest.osgi/src/main/java/com/ericsson/bss/cassandra/ecchronos/rest/osgi/RepairManagementRESTComponent.java
```diff
@@ -64,9 +64,9 @@ public String tableStatus(String keyspace, String table)
     }
 
     @Override
-    public String jobStatus(String keyspace, String table, String id)
+    public String jobStatus(String keyspace)
```
emolsson, line 67: id

rest/src/main/java/com/ericsson/bss/cassandra/ecchronos/rest/RepairManagementRESTImpl.java
itskarlsson: The request at the moment sends back an empty result if the id doesn't match an id in the system. It does currently not differentiate between bad formatting or incorrect id.

The standard use case of this request is to fetch a repair job whose id you have just read(with the cli for example). Never will you guess an id or come up with one on the fly (Which is the most common way to end up with a malformed UUID)

Therefore, the only distinction that would give value is whether or not something exists or not.

emolsson (approved):

PR #180, opened 2021-02-18
Step mockito-core to 3.3.3

ArcturusMengsk:
Plus some related fixes.

### DIFF
--- a/connection/pom.xml
+++ b/connection/pom.xml
@@ -54,7 +54,7 @@
 
         <dependency>
             <groupId>org.mockito</groupId>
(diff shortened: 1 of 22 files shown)


itskarlsson:
Code LGTM.

One comment. Ensure that the pull request is targeted towards the minimum possible version. This is something that lower versions can benefit from. Just a quick check on my part showed that this merges cleanly with 1.2. Could you check the viability/need of getting this patch into lower versions?

itskarlsson (approved):

PR #199, opened 2021-04-20
Fix hash bug when restarting ondemand jobs

tommystendahl:


itskarlsson (approved):
LGTM, just change commit message to: 'Fix hash not calculated correctly for on demand repair jobs'

Issue #172, opened 2021-01-29
On-demand repair jobs not started after Cassandra restart


etedpet:
When a Cassandra nod is restarted, the following is printed in the ecChronos log:

> {"timestamp":"2021-01-29T10:27:11.824Z","severity":"warning","message":"Failed to check connection: java.net.ConnectException: Connection refused (Connection refused)"}
> {"timestamp":"2021-01-29T10:27:11.825Z","severity":"warning","message":"stopping"}
> {"timestamp":"2021-01-29T10:27:11.970Z","severity":"error","message":"Unable to update table sizes, future metrics might contain stale data"}

After that it is no longer possible to trigger an on-demand repair. 

Example with the (ks1.tb1 table):
```
$ curl -X POST http://localhost:8081/repair-management/v1/schedule/keyspaces/ks1/tables/tb1
{"keyspace":"ks1","table":"tb1","lastRepairedAtInMs":-1,"repairedRatio":0.0,"status":"IN_QUEUE","nextRepairInMs":-1,"id":"c72be6e6-f5d8-44df-8cf3-61180f44903c","recurring":false}
```

The request is accepted and a response with an Id is returned. But nothing is printed to the logs and the repair on ks1.tbt is not triggered.

The following status is shown:
```
$ ecc-status -k ks1
---------------------------------------------------------------------------------------------------------------------------------------------
| Id                                   | Keyspace | Table | Status    | Repaired(%) | Repaired at         | Next repair         | Recurring | 
---------------------------------------------------------------------------------------------------------------------------------------------
| c72be6e6-f5d8-44df-8cf3-61180f44903c | ks1      | tb1   | IN_QUEUE  | 0.00        | -                   | -                   | False     |   <--- "Triggered on-demand repair"
| 9750dc30-621b-11eb-ad6c-c189d374d30c | ks1      | tb1   | COMPLETED | 100.00      | 2021-01-23 10:20:21 | 2021-01-30 10:20:21 | True      | 
---------------------------------------------------------------------------------------------------------------------------------------------
```

But the job is stuck in `IN_QUEUE` and never starts.

**If** ecChronos is also restarted, then the on-demand repair fails (`ERROR`), BUT another repair can then be triggered!

```
$ curl -X POST http://localhost:8081/repair-management/v1/schedule/keyspaces/ks1/tables/tb1
[...]

PR #206, opened 2021-05-07
Handle if we get empty token map after C* restart

tommystendahl:
Closes issue #172


itskarlsson (commented):
LGTM. However, I think we should add one or more unit tests here.

itskarlsson (approved):

PR #214, opened 2021-06-14
Mitigate pull limit issues

itskarlsson:


itskarlsson:
Not doing this as we are going to remove travis

PR #242, opened 2021-11-22
Add EndPointFactory

emolsson:
* Treat endpoints as the same endpoint regardless of IP as long
  as the host UUID is the same.
* Add a contact endpoint for the case when a resolvable hostname
  is used for the local node. This will resolve to the specified
  hostname and always return the same endpoint object when the
  factory asks for it.

### DIFF
--- a/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairSchedulerImpl.java
+++ b/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairSchedulerImpl.java
@@ -128,9 +128,16 @@ public class RepairSchedulerImpl implements RepairScheduler, Closeable
     {
         synchronized (myLock)
         {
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/RepairSchedulerImpl.java
```diff
@@ -128,9 +128,16 @@ private void handleTableConfigurationChange(TableReference tableReference, Repai
...
         {
-            if (configurationHasChanged(tableReference, repairConfiguration))
+            try
             {
-                createTableSchedule(tableReference, repairConfiguration);
+                if (configurationHasChanged(tableReference, repairConfiguration))
+                {
+                    createTableSchedule(tableReference, repairConfiguration);
+                }
+            }
```
itskarlsson, line 137: I suppose you are pushing the fixes i did for #241

tommystendahl (approved):

PR #244, opened 2021-12-05
Fix flakiness in some tests

pthariensflame:
These tests were failing under [NonDex](https://github.com/TestingResearchIllinois/NonDex) due to implementation-defined behavior of collections; this PR ensures that these tests ignore order when irrelevant, and preserve order when necessary.


itskarlsson (commented):
Thanks for the contribution

Issue #158, opened 2020-10-30
(On-demand) Repair scheduling tool


emolsson:
Currently we are lacking the possibility to schedule repairs using a tool, rather than using pure REST requests.
It's also not possible to schedule repairs for all keyspaces or all tables within a keyspace.

Items to complete:
* Repair multiple tables in a single command (REST or tool based).
* Add a python tool for scheduling repairs.
* Make sure we use the correct configuration when repairing tables on-demand.

etedpet:
Also add the possibility to specify that only the _primary replicas_ of the data should be repaired. 
Similar to `nodetool repair --partitioner-range` which is recommended for routine maintenance.

This way the on-demand repair can be triggered on all nodes (for repairing a table or the entire cluster on demand) without having to repair everything x 3 (given RF=3)

PR #251, opened 2022-01-18
ecc_schedule command

masokol:


itskarlsson:
fixes #158

ecchronos-binary/src/bin/ecc_schedule.py
```diff
@@ -0,0 +1,70 @@
...
+def print_repair_job(repair_job):
+    repair_jobs_table = [["Id", "Keyspace", "Table", "Status", "Repaired(%)",
+                          "Completed at", "Next repair", "Recurring"]]
+    repair_jobs_table.append(convert_repair_job(repair_job))
+    table_formatter.format_table(repair_jobs_table)
+
+
+def parse_arguments():
+    parser = ArgumentParser(description='Schedule a repair')
+    parser.add_argument('-k', '--keyspace', type=str,
```
itskarlsson, line 49: There are a few other options that might be interesting here.
For example, partitioner--range as noted in #158.
Would be interesting to add but would require a far more invasive change. Certainly an improvement for the future to think about.

itskarlsson (approved):
Add an entry in the changelog and make sure you format the commit message correctly. 
Otherwise, LGTM

ecchronos-binary/src/bin/ecc_schedule.py
masokol: Agree, next up is merging the commands. After that i will see if i can fix all the issues mentioned in #158

PR #279, opened 2022-03-08
Offset next runtime with repair time taken

### LINKED ISSUE #121: Calculate previous run-time to determine when to run next
Currently the repair scheduler is dynamically shifting around the schedule based on when ranges where last repaired.
While this is good for spreading the schedule there are still some inaccuracy with regards to repair interval targets. The current approach calculates `"start of repair of range" + repair interval` as the next time we should start a repair which causes a constant shift and misses the target _every time_.

Since we have both a start and finish time per range we could instead estimate how long the repair will take and subtract that from the "next start" to make the next start be `"start of repair of range" + repair interval - estimated repair time` instead.

On a table level this needs to be a summarized estimated repair time for all ranges that should be repaired.

---

While this won't solve the issue of estimating repair time between tables (tableX and tableY have the same interval and want to start at roughly the same time) that should in theory be shifted enough over time that there no longer is an overlap between the two tables.

masokol:
Closes #121


itskarlsson (commented):
Besides this, we should probably test what will happen if the estimated time is higher than the interval.

Also, it would be cool if we could get statistics on how much 'effective repair time' remains per job.

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/metrics/TableRepairMetrics.java
```diff
@@ -40,6 +40,13 @@
      */
     void lastRepairedAt(TableReference tableReference, long lastRepairedAt);
 
+    /**
+     * Report the remaining repair time for table.
+     * @param tableReference The table to update the remaining repair time for.
+     * @param remainingRepairTime The remaining time to fully repair the table.
```
itskarlsson, line 46: Probably good to mention that this is active repair time.
Aka the time spend waiting for cassandra to repair as opposed to the time for a job to complete.

itskarlsson (commented):
The remaining repair would be good to present in ecctool but this could be a future improvement

tommystendahl (approved):

masokol:
> The remaining repair would be good to present in ecctool but this could be a future improvement

#283

masokol:
Changelog will be updated once #288 is merged

tommystendahl (approved):

PR #287, opened 2022-03-30
Add blocked status for jobs blocked by run policy

### LINKED ISSUE #284: Add specific status for repair jobs when they are blocked by run policy
It would be good to include halt/blocked status for repair jobs when they're blocked by [TimeBasedRunPolicy](https://github.com/Ericsson/ecchronos/blob/master/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/TimeBasedRunPolicy.java). Currently repair jobs will have status IN_QUEUE and the only way to see that they're blocked is to look at debug logs.

masokol:
Closes #284

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Version 3.0.0
 
+* Add blocked status - Issue #284
+
 ## Version 2.0.6
 
 * Add ecc-schedule command - Issue #158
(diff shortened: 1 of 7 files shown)


masokol:
Changelog will be updated once #288 is merged

tommystendahl (approved):

PR #302, opened 2022-05-02
Support for cluster-wide repairs

### LINKED ISSUE #299: Cluster Wide Repair
Currently, on demand repair must be triggered from all nodes to repair the full cluster. There is no room to specify -pr which leads to data being repaired multiple times.

We should allow for cluster wide repair by triggering from one instance an propagating the repair over the cluster.

The two issues that this would solve are these:
1. Repair needs to be triggered on all nodes
2. The same data is repaired multiple times

One possible solution discussed is creating RepairGroups and instruct other nodes to do the same. The repair could keep track of which RepairGroups have been created and only create missing groups to repair each RepairGroup once.

masokol:
Closes #299


core/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/repair/TestOnDemandStatus.java
```diff
@@ -251,6 +251,42 @@ public void testUpdateToFailed()
...
+    @Test
+    public void testGetAllClusterWideJobs()
+    {
+        OnDemandStatus onDemandStatus = new OnDemandStatus(getNativeConnectionProvider());
+
+        UUID jobId = UUID.randomUUID();
+        int hashValue = 1;
+        TableReference tableReference = myTableReferenceFactory.forTable(KEYSPACE_NAME, TEST_TABLE_NAME);
+        Map<LongTokenRange, ImmutableSet<Node>> tokenMap = new HashMap<>();
+        when(myReplicationState.getTokenRangeToReplicas(tableReference)).thenReturn(tokenMap );
```
itskarlsson, line 273: Remove space (tokenMap **)**

ecchronos-binary/src/bin/ecctool.py
```diff
@@ -84,6 +84,8 @@ def add_repairs_subcommand(sub_parsers):
     parser_repairs.add_argument("-l", "--limit", type=int,
                                 help="Limit the number of tables or virtual nodes printed (-1 to disable)",
                                 default=-1)
+    parser_repairs.add_argument("--hostid", type=str,
+                                help='Show repairs for the speficied host id')
```
itskarlsson, line 88: specified

ecchronos-binary/src/bin/ecctool.py
```diff
@@ -135,6 +137,8 @@ def add_run_repair_subcommand(sub_parsers):
     parser_trigger_repair.add_argument("-u", "--url", type=str,
                                        help="The host to connect to with the format (http://<host>:port)",
                                        default=None)
+    parser_trigger_repair.add_argument("--local", action='store_true',
+                                       help='repair will run for the local node, default is False', default=False)
```
itskarlsson, line 141: disabled by default (since there is no true false from the users perspective)

ecchronos-binary/src/pylib/ecchronoslib/rest.py
```diff
@@ -186,13 +187,19 @@ def list_schedules(self, keyspace=None, table=None):
...
             request_url = "{0}?keyspace={1}&table={2}".format(request_url, keyspace, table)
+            if host_id:
+                request_url += "&hostId={0}".format(host_id)
         elif keyspace:
             request_url = "{0}?keyspace={1}".format(request_url, keyspace)
+            if host_id:
+                request_url += "&hostId={0}".format(host_id)
+        else:
+            if host_id:
+                request_url += "?hostId={0}".format(host_id)
```
itskarlsson, line 203: Couldn't all of this be just a single if statement afterwards?

standalone-integration/src/test/java/com/ericsson/bss/cassandra/ecchronos/standalone/ITOnDemandRepairJob.java
```diff
@@ -186,7 +184,7 @@ public void clean()
         reset(mockTableRepairMetrics);
     }
 
-    @AfterClass
+    @Parameterized.AfterParam
```
itskarlsson, line 187: What is the effect of this change?

ecchronos-binary/src/test/behave/features/steps/ecctool_get_repairs_steps.py
```diff
@@ -75,6 +75,12 @@ def step_show_repair(context, keyspace, table):
     run_ecc_repair_status(context, ['--keyspace', keyspace, '--table', table])
     handle_repair_output(context)
 
+@when(u'we list repairs for hostid and table {keyspace}.{table}')
+def step_show_repair_with_nodeid(context, keyspace, table):
+    run_ecc_repair_status(context, ['--keyspace', keyspace, '--table', table,
+                                    '--hostid', '{0}'.format(context.host_id)])
```
itskarlsson, line 81: It might be good to give this value (and all other values set in environment.py) another name. Something that will enforce that this is set in the environment file. Like context.environment.host_id or similar.

itskarlsson (changes requested):

ecchronos-binary/src/pylib/ecchronoslib/rest.py
masokol: nope, if keyspace or table is set we need to append &hostid, if none are specified we need to append ?hostid

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/OnDemandRepairSchedulerImpl.java
```diff
@@ -68,33 +72,20 @@ private OnDemandRepairSchedulerImpl(Builder builder)
...
-                {
-                    Thread.currentThread().interrupt();
-                }
-            }
+            Set<OngoingJob> ongoingJobs = myOnDemandStatus.getOngoingJobs(myReplicationState);
+            ongoingJobs.forEach(j -> scheduleOngoingJob(j));
+        }
+        catch (Exception e)
+        {
+            LOG.info("Failed to get ongoing ondemand jobs: {}, automatic retry in {}s", e.getMessage(),
```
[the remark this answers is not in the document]
masokol: it's scheduled at fixed rate, so it will have automatic retry.

standalone-integration/src/test/java/com/ericsson/bss/cassandra/ecchronos/standalone/ITOnDemandRepairJob.java
masokol: Things are initialized in method above with "@beforeparam", that's why we should have "afterparam" here otherwise we will leak resources each time beforeparam is called.

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/ScheduledRepairJobView.java
```diff
@@ -47,4 +47,10 @@ public Boolean isRecurring()
     {
         return true;
     }
+
+    @Override
+    public UUID getHostId()
+    {
+        return null;
```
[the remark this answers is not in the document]
masokol: I've made it a bit better, we will be able to split them when we remove v1 REST interface because that's where they are combined today.

PR #305, opened 2022-05-13
Make fault reporter pluggable

masokol:


application/src/main/resources/ecc.yml
```diff
@@ -103,6 +103,11 @@ repair:
   ## if the table has not been properly repaired within one full day.
   ##
   alarm:
+    ##
+    ## The class used for fault reporting
+    ## The default LoggingFaultReporter will log when alarm is raised/ceased
```
itskarlsson, line 108: NIT: 'The default LoggingFaultReporter will log when **an** alarm is raised/ceased'

PR #307, opened 2022-05-20
Add cassandra health indicator and enable probes

### LINKED ISSUE #192: Add liveness endpoint to REST API
The following REST endpoint is provided today (by the Spring framework?)

> http://< host >:< port >/actuator/health

Does that guarantee that the ecChronos application is live (working as expected)? 
Or is a separate endpoint provided by ecChronos needed, to verify that the repair scheduling is working as expected.

It should be possible to use this REST endpoint as an liveness-probe in a k8s cluster.

masokol:
actuator/health/readiness will report readiness of ecchronos.
The above will check CQL and JMX connection towards cassandra.
actuator/health/liveness includes generic springboot liveness.

Closes #192


masokol:
We might want to include more checks if someone has any ideas what would be good to have.

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/spring/CassandraHealthIndicator.java
```diff
@@ -0,0 +1,104 @@
...
+                    .withJmxConnectionProvider(myJmxConnectionProvider)
+                    .build();
+            JmxProxy jmxProxy = jmxProxyFactory.connect();
+            jmxProxy.close();
+            return true;
+        }
+        catch (Exception e)
+        {
+            LOG.debug("JMX readiness failed", e);
+            LOG.error("JMX readiness failed {}", e.getMessage());
```
itskarlsson, line 78: Not so sure about using readiness terminology in our healthcheck. 

"Unable to connect over JMX" - I would like better.

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/spring/CassandraHealthIndicator.java
```diff
@@ -0,0 +1,104 @@
...
+            Session session = myNativeConnectionProvider.getSession();
+            ResultSet result = session.execute(selectFromLocal);
+            if (!result.all().isEmpty())
+            {
+                return true;
+            }
+        }
+        catch (Exception e)
+        {
+            LOG.debug("CQL readiness failed", e);
```
itskarlsson, line 98: Not so sure about using readiness terminology in our healthcheck.

"Unable to connect over CQL" - I would like better.

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/spring/CassandraHealthIndicator.java
```diff
@@ -0,0 +1,104 @@
...
+            LOG.error("CQL readiness failed {}", e.getMessage());
```
itskarlsson, line 99: Also not a big fan of double printing error messages. Do we need the error message here?

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/spring/CassandraHealthIndicator.java
itskarlsson, line 78: Also not a big fan of double printing error messages. Do we need the error message here?

application/src/main/resources/application.yml
```diff
@@ -24,6 +24,13 @@ management:
   endpoint:
     health:
       enabled: true
+      probes:
+        enabled: true
```
itskarlsson, line 28: As I understand it, these probes are enabled by default according to (https://www.baeldung.com/spring-liveness-readiness-probes) when using a Kubernetes environment. Why do we enable it for all environments?

masokol: I don't trust automatic detection, "automatic detection" usually doesn't work 100% of the time. Also, i think it's nice to have alive/ready split up.

itskarlsson (approved):

PR #308, opened 2022-05-23
Fix malformed IPv6 for jmx in new Java versions

### LINKED ISSUE #306: Malformed IPv6 address during startup
Since [JDK-8278972](https://www.oracle.com/java/technologies/javase/8u331-relnotes.html#JDK-8278972) it's not possible to start EcChronos when IPv4 addresses are used.

Inspiration for how to solve it can be taken from a similar Cassandra ticket [CASSANDRA-17581](https://issues.apache.org/jira/browse/CASSANDRA-17581).

Workaround:
Add the line:
```
-Dcom.sun.jndi.rmiURLParsing=legacy
```
to `jvm.options` and this will disable the new validation.

masokol:
Closes #306

### DIFF
--- a/connection.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/connection/impl/LocalJmxConnectionProvider.java
+++ b/connection.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/connection/impl/LocalJmxConnectionProvider.java
@@ -34,7 +34,7 @@ public class LocalJmxConnectionProvider implements JmxConnectionProvider
 {
     private static final Logger LOG = LoggerFactory.getLogger(LocalJmxConnectionProvider.class);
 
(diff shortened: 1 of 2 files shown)


connection.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/connection/impl/LocalJmxConnectionProvider.java
```diff
@@ -84,7 +84,13 @@ public void close() throws IOException
 
     private void reconnect() throws IOException
     {
-        JMXServiceURL jmxUrl = new JMXServiceURL(String.format(JMX_FORMAT_URL, myLocalhost, myPort));
+        String host = myLocalhost;
+        if (host.contains(":"))
+        {
+            // Use square brackets to surround IPv6 addresses to fix
```
itskarlsson, line 90: NIT: It should be enough with:
Use square brackets to surround IPv6 addresses

itskarlsson (approved):

PR #325, opened 2022-06-28
 Ongoing repairs sorted to the top

### LINKED ISSUE #314: New repairs are sorted to be last in ecctool
When using ecctool to fetch information ongoing repairs. The repairs which are currently running are display last when they should be first.
The reason for this is that we sort after completed time and non completed repairs have a -1 in that field.

jwaeab:
…ongoing repairs are at the top of the list. This will make limit more useful as well. (#314)

Fixes #314

### DIFF
--- a/ecchronos-binary/src/pylib/ecchronoslib/table_printer_v2.py
+++ b/ecchronos-binary/src/pylib/ecchronoslib/table_printer_v2.py
@@ -89,8 +89,8 @@ def print_repairs(repairs, max_lines=-1):
 
 def print_schedule_table(schedule_table, schedules, max_lines):
 
-    sorted_schedules = sorted(schedules, key=lambda sorted_schedule: sorted_schedule.last_repaired_at_in_ms,
-                              reverse=True)
+    sorted_schedules = sorted(schedules, key=lambda x: (x.last_repaired_at_in_ms, x.repaired_ratio),
+                              reverse=False)
     if max_lines > -1:
         sorted_schedules = sorted_schedules[:max_lines]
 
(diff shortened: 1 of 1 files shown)


ecchronos-binary/src/pylib/ecchronoslib/table_printer_v2.py
```diff
@@ -89,8 +89,8 @@ def print_repairs(repairs, max_lines=-1):
 
 def print_schedule_table(schedule_table, schedules, max_lines):
 
-    sorted_schedules = sorted(schedules, key=lambda sorted_schedule: sorted_schedule.last_repaired_at_in_ms,
-                              reverse=True)
+    sorted_schedules = sorted(schedules, key=lambda x: (x.last_repaired_at_in_ms, x.repaired_ratio, x.status),
```
masokol, line 92: I would prefer if status wasn't part of the sorting. It's a string so it's hard to predict how sorting will be performed.

ecchronos-binary/src/pylib/ecchronoslib/table_printer_v2.py
```diff
@@ -100,7 +100,7 @@ def print_schedule_table(schedule_table, schedules, max_lines):
 
 def print_repair_table(repair_table, repairs, max_lines):
 
-    sorted_repairs = sorted(repairs, key=lambda sorted_repair: sorted_repair.completed_at, reverse=True)
+    sorted_repairs = sorted(repairs, key=lambda x: (x.completed_at, x.repaired_ratio, x.status), reverse=False)
```
masokol, line 103: same as above.

ecchronos-binary/src/pylib/ecchronoslib/table_printer_v2.py
jwaeab: I'm not sure I follow this one. Sorting will be in descending order. The status is somewhat hardcoded throughout, so we couldn't just change it like that anyhow.

ecchronos-binary/src/pylib/ecchronoslib/table_printer_v2.py
jwaeab: Same as above.

masokol (approved):

PR #336, opened 2022-08-16
Listen for keyspace creations to create schedules

masokol:
Only listening on table creations is not enough, onKeyspaceCreated is
called instead if keyspace and table are created before the schema
agreement completes. Therefore, listening on both keyspace and table
creations is mandatory to not miss events.


jwaeab (commented):
Ok

tommystendahl (approved):

PR #378, opened 2022-09-26
Add support for certificates in ecctool

### LINKED ISSUE #374: Support mTLS in ecctool
Currently ecctool does not support mTLS (mutual TLS).
Also it would be good for the python integration tests to be updated to use mTLS.

masokol:
Closes #374


cassandra-test-image/src/main/resources/generate_certificates.sh
```diff
@@ -26,6 +27,19 @@
...
 #
+# Certs for REST
+# cert/serverca.crt - REST CA
+# cert/servercakey.pem - REST CA private key
+# cert/server.csr - REST server CSR
+# cert/servercert.crt - REST server cert
+# cert/serverkey.pem - REST server private key
+# cert/serverkeystore - REST server keystore (PKCS12)
+# cert/servertruststore - REST server truststore (PKCS12)
+#
```
itskarlsson, line 38: NIT:
\# Certs for ecctool

ecchronos-binary/src/pylib/ecchronoslib/rest.py
```diff
@@ -77,7 +79,15 @@ def request(self, url, method='GET'):
...
             request = Request(request_url)
             request.get_method = lambda: method
-            response = urlopen(request)
+            cert_file = os.getenv("ECCTOOL_CERT_FILE")
+            key_file = os.getenv("ECCTOOL_KEY_FILE")
+            ca_file = os.getenv("ECCTOOL_CA_FILE")
+            if cert_file and key_file and ca_file:
+                context = ssl.create_default_context(cafile=ca_file)
+                context.load_cert_chain(cert_file, key_file)
+                response = urlopen(request, context=context)
```
itskarlsson, line 88: Will the request not be done with http by default? Wouldn't this mean that the user will always have to specify url if he uses certificates? This does not seem right.

itskarlsson (approved):

PR #394, opened 2022-10-17
Add possibility to decide reporting for metrics

### LINKED ISSUE #390: Add possibility to turn off writing metrics to file but keep producing prometheus metrics
Currently metrics are either disabled or enabled but there are some use cases that would like to only do one or the other.

masokol:
Closes #390

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,8 @@
 
 ## Version 4.0.0
 
-* Disable driver metrics if statistics.enabled=false - #391
+* Add possibility to decide reporting for metrics - Issue #390
+* Disable driver metrics if statistics.enabled=false - Issue #391
 * Reload sslContext for cql client if certs change - Issue #329
 * Expose java driver metrics - Issue #368
 * Split metrics into separate connector - Issue #369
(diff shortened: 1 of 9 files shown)


CHANGES.md
```diff
@@ -2,7 +2,8 @@
 
 ## Version 4.0.0
 
-* Disable driver metrics if statistics.enabled=false - #391
+* Add possibility to decide reporting for metrics - Issue #390
+* Disable driver metrics if statistics.enabled=false - Issue #391
```
itskarlsson, line 6: Should be first

masokol: This is 390, the 391 is already merged.

application/src/main/resources/ecc.yml
```diff
@@ -167,6 +167,16 @@ repair:
 
 statistics:
   enabled: true
+  ##
+  ## How metrics should be reported
```
itskarlsson, line 171: Some comment about what is the behavior if all are disabled but statistics is true would be good here. As well as if statistics is false and all are enabled.

application/src/test/resources/all_reporting_disabled.yml
```diff
@@ -0,0 +1,25 @@
+#
+# Copyright 2020 Telefonaktiebolaget LM Ericsson
```
itskarlsson, line 2: 2022

itskarlsson (approved):

PR #403, opened 2022-10-18
Able to override ReloadingCertificateHandler

### LINKED ISSUE #399: ReloadingCertificateHandler cannot be overriden
Methods in ReloadingCertificateHandler have been made final, it's not longer possible to override them.
We have users that do that, so this should be fixed.

ch1bbe:
Closes #399


masokol:
LGTM

PR #413, opened 2022-10-26
Add name field to each maven module

### LINKED ISSUE #371: Populate <name> field in pom.xml files
Hi,

This is more of a request than an actual issue... But, automation that we are using is relying on parsing the pom.xml files, and seems to be using the <name> field. When this field is not populated, it falls back to <artifactId>. Looking at these in ecChronos, they are all pretty "generic" so to speak, like "application", "core", "fm", and so on. Therefore, they are not very useful.

What I would like is for the <name> fields in each pom.xml to be populated with slightly more descriptive names. If we take the examples above, it would be as easy as having:
 * ecChronos Application
 * ecChronos Core
 * ecChronos Fault Management

Here's another random component to illustrate:
https://github.com/eclipse-ee4j/jaf/blob/master/api/pom.xml

masokol:
Closes #371

### DIFF
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -25,6 +25,7 @@
 
     <artifactId>application</artifactId>
     <description>Standalone application of ecChronos</description>
+    <name>ecChronos application</name>
 
     <dependencies>
         <!-- Internal -->
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -25,7 +25,7 @@
 
     <artifactId>application</artifactId>
     <description>Standalone application of ecChronos</description>
-    <name>ecChronos application</name>
+    <name>EcChronos Application</name>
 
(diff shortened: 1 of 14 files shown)


ch1bbe (approved):

itskarlsson:
@ArcturusMengsk LGTY?

ArcturusMengsk:
I would prefer title capitalization, if possible. According to https://capitalizemytitle.com/, **Chicago** style.
But if it's already merged, then don't bother.

masokol:
@ArcturusMengsk Is this what you had in mind?

ch1bbe (approved):

ArcturusMengsk:
> @ArcturusMengsk Is this what you had in mind?

Yes, thank you!

PR #419, opened 2022-10-31
Allow global prefix for metrics

### LINKED ISSUE #417: Add support for prefixing all metrics
It would be nice if all metrics produced by ecchronos would be prefixed.
It would make finding metrics much easier in for example prometheus.

The prefix should be configurable in ecc.yml.

masokol:
Closes #417

### DIFF
--- a/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
+++ b/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
@@ -76,6 +76,7 @@ public class DefaultNativeConnectionProvider implements NativeConnectionProvider
                 .withSslEngineFactory(sslEngineFactory)
                 .withMetricRegistry(metricRegistry)
                 .withMetricsEnabled(config.getStatistics().isEnabled())
+                .withMetricPrefix(config.getStatistics().getPrefix())
                 .withSchemaChangeListener(defaultRepairConfigurationProvider);
 
         myLocalNativeConnectionProvider = establishConnection(nativeConnectionBuilder,
--- a/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/ECChronosInternals.java
+++ b/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/ECChronosInternals.java
@@ -109,6 +109,7 @@ public class ECChronosInternals implements Closeable
                     .withMetricRegistry(metricRegistry)
                     .withJmxReporting(configuration.getStatistics().getReporting().isJmxReportingEnabled())
                     .withFileReporting(configuration.getStatistics().getReporting().isFileReportingEnabled())
+                    .withMetricPrefix(configuration.getStatistics().getPrefix())
                     .build();
         }
         else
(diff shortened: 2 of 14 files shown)


jwaeab (approved):

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
```diff
@@ -76,6 +76,7 @@ public DefaultNativeConnectionProvider(final Config config,
                 .withSslEngineFactory(sslEngineFactory)
                 .withMetricRegistry(metricRegistry)
                 .withMetricsEnabled(config.getStatistics().isEnabled())
+                .withMetricPrefix(config.getStatistics().getPrefix())
```
itskarlsson, line 79: Perhaps it could be good to refactor some of these into just one thing. So you are just passing statistics.

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/ECChronosInternals.java
```diff
@@ -109,6 +109,7 @@ public ECChronosInternals(final Config configuration,
                     .withMetricRegistry(metricRegistry)
                     .withJmxReporting(configuration.getStatistics().getReporting().isJmxReportingEnabled())
                     .withFileReporting(configuration.getStatistics().getReporting().isFileReportingEnabled())
```
itskarlsson, line 111: These might also be able to be refactored so just statistics needs to be passed

docs/METRICS.md
```diff
@@ -23,6 +23,17 @@ can be excluded in the same way as ecChronos metrics.
...
 
+## Metric prefix
+
+It's possible to define a global prefix for all metric names.
+This is done by specifying a string in `statistics.prefix` in `ecc.yml`.
+The prefix cannot start or end with a dot or any other path separator.
+
+For example if the prefix is `ecChronos` and the metric name is `RepairSuccessTime`,
+the metric name will be `ecChronos.RepairSuccessTime`.
+
```
itskarlsson, line 34: Would be good to describe how this prepends on the different types of reporting

itskarlsson (changes requested):

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
masokol: Statistics is in application module, application depends on core. We can't use statistics in core because that will lead to circular dependency.

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/ECChronosInternals.java
masokol: Statistics is in application module, application depends on core, connection etc. We can't use statistics in core because that will lead to circular dependency.

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
itskarlsson: It would still be good to pair metrics into it's own object since we are sending it into a Connection but I am fine without it.

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/ECChronosInternals.java
itskarlsson: It would still be good to pair metrics into it's own object since we are sending it into a Connection but I am fine without it.

itskarlsson (approved):

PR #506, opened 2023-06-12
Fix ecctool shebang

### LINKED ISSUE #504: Fix shebangs in ecctool
shebangs used in ecctool are wrong, we should use /usr/bin/env python3

masokol:
Closes #504

### DIFF
--- a/ecchronos-binary/src/bin/ecctool.py
+++ b/ecchronos-binary/src/bin/ecctool.py
@@ -1,5 +1,4 @@
-#!/bin/sh
-""""exec python -B -- "$0" ${1+"$@"} # """
+#!/usr/bin/env python3
 # vi: syntax=python
 #
 # Copyright 2022 Telefonaktiebolaget LM Ericsson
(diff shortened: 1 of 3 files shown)


itskarlsson:
I have tested it and it seems to solve the issue.
The issue being that the lack of soft-link to /usr/bin/python causes the python tests to fail.

However, this issue is even in older versions and should be fixed there as well.

ecchronos-binary/src/bin/ecctool.py
```diff
@@ -1,5 +1,4 @@
-#!/bin/sh
-""""exec python -B -- "$0" ${1+"$@"} # """
+#!/usr/bin/env python3
```
itskarlsson, line 1: Probably need to target an older version and merge forward

masokol: I don't think we need to since it's now we only support python3. Previously we supported both python 2 and python 3 and the soft-link python always pointed to python 2.

tommystendahl (approved):

ecchronos-binary/src/bin/ecctool.py
itskarlsson: Sounds reasonable

itskarlsson (approved):

PR #516, opened 2023-06-21
Fix priority for repair jobs

### LINKED ISSUE #515: Prioritization of scheduled jobs is not correct while Cassandra nodes are considered down
Prioritizaion of TableRepairJob and tasks it creates (RepairGroup) is not correct.
The real priority of TableRepairJob is based on vnodes that will not be repaired because some of the replicas are down:
See calculation of repairedAt and excluding [here](https://github.com/Ericsson/ecchronos/blob/9037f507e04a3247d31b990d848a55ea6d4a047d/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/state/RepairStateImpl.java#L124)

More or less what happens is that, we calculate repairedAt based on all vnodes and replicas, later on we exclude some vnodes because replicas are down. The repairedAt is used for calculating realPriority, i.e the priority might be higher for jobs than intended. The consequences of this is too high priority in lock_priority table which slows down repairs because ecchronos won't even try to lock if the local priority is lower than what is in that table.

I think a solution to this would be to calculate the priority based on what the job can repair.

masokol:
Closes #515

### DIFF
--- a/core/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/repair/TestTableRepairJob.java
+++ b/core/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/repair/TestTableRepairJob.java
@@ -22,21 +22,30 @@ import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.ignoreStubs;
+import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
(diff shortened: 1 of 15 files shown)


tommystendahl (approved):

core/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/repair/TestTableRepairJob.java
```diff
@@ -477,6 +495,7 @@ public void testLastSuccessfulRunIsBasedOnRepairHistory()
         // Repair has been completed
         doReturn(lastRepairedAtAfterRepair).when(myRepairStateSnapshot).lastRepairedAt();
         doReturn(false).when(myRepairStateSnapshot).canRepair();
```
masokol, line 497: Maybe add some more tests for the priority method.

itskarlsson (approved):

PR #528, opened 2023-08-01
Refactor config package

### LINKED ISSUE #488: Investigate if refactoring some parts is necessary
Although some cleanup was done in #480, there are still places (with code style suppression) which needs to be cleaned up or investigated further if refactoring is necessary.

masokol:
Partially fixes #488


masokol:
Unittests will be split up in a separate PR (this PR is already big).

tommystendahl (approved):

PR #547, opened 2023-08-14
Fix priority calculation for local queue

### LINKED ISSUE #546: ecChronos wrong priority calculation for local queue
ecChronos has a local queue of repair jobs that should be run based on the priority. RepairState used to calculate the priority might be stale because it's updated first when job is tried to be run from the queue. More or less the problem is that if repair jobs before take longer time than repair interval to be run then repair jobs in the end of the queue will never get their states updated because the priority for newly run jobs will be higher.

Example (first run)
Queue:
R1 (priority -1)
R2 (priority -1)
R3 (priority -1)

If R1 takes longer time than interval then the queue will look like this:
R1 (priority 1)
R2 (priority -1)
R3 (priority -1)

Instead of defaulting the priority to -1 when there's nothing to repair according to repairState, calculate the priority based on lastRepairedAt (as it was pre #515). By doing so, all the jobs in the local queue will be tried to run, effectively updating the repairState and solving the problem.

With the fix the queue will look like this:

R1 (priority -1)
R2 (priority -1)
R3 (priority -1)

After R1 has run, X > 0.
R2 (priority X) 
R3 (priority X)
R1 (priority 1)

masokol:
Closes #546

### DIFF
--- a/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/TableRepairJob.java
+++ b/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/TableRepairJob.java
@@ -69,7 +69,6 @@ public class TableRepairJob extends ScheduledJob
         myRepairConfiguration = builder.repairConfiguration;
         myRepairLockType = builder.repairLockType;
         myRepairPolicies = builder.repairPolicies;
-        refreshState();
     }
 
     public TableReference getTableReference()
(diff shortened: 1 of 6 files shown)


tommystendahl (approved):

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/repair/TableRepairJob.java
```diff
@@ -208,7 +208,8 @@ public boolean runnable()
     public final int getRealPriority()
     {
         RepairStateSnapshot repairStateSnapshot = myRepairState.getSnapshot();
-        int priority = -1;
+        // Default to lastRepairedAt, unless we have repairGroups that we can repair.
```
masokol, line 211: This is not 100% correct either. If we have a repairGroup that can run now, the priority in the queue will be based on that. If a repairGroup becomes available for run a few minutes later then it will have lower priority than it should have locally in the queue.

masokol: I will need to rethink this, the ideal would be to always have the repairState up to date and base priority on that.

tommystendahl (approved):

emolsson (approved):

PR #568, opened 2023-09-06
Set heap size when running python ITs

### LINKED ISSUE #539: Python integration tests should be running with jvm.options
It would be good if we would configure heap size for ecChronos when running in python-integration-tests.

masokol:
Closes #539

### DIFF
--- a/ecchronos-binary/src/test/bin/pytests.sh
+++ b/ecchronos-binary/src/test/bin/pytests.sh
@@ -42,6 +42,8 @@ CERTIFICATE_DIRECTORY=${project.build.directory}/certificates/cert
 ## Connection
 # Remove comment lines
 sed '/^\s*#.*/d' -i "$CONF_DIR"/ecc.yml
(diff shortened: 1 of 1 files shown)


ecchronos-binary/src/test/bin/pytests.sh
```diff
@@ -42,6 +42,8 @@ CERTIFICATE_DIRECTORY=${project.build.directory}/certificates/cert
 ## Connection
 # Remove comment lines
 sed '/^\s*#.*/d' -i "$CONF_DIR"/ecc.yml
+# Uncomment heap options
+sed 's/^#-X/-X/g' -i "$CONF_DIR"/jvm.options
```
jwaeab, line 46: Would it be better to consume whitespaces as well, such as in 's/^#\s*-X/-X/g'?

jwaeab (approved):

tommystendahl (approved):

PR #615, opened 2023-11-17
OpenSSF best practices: set GitHub token permission to read-only

gkunz:


gkunz:
Hi @tommystendahl. I just noticed that you seem to push to branches instead of to master directly as I did in this PR. I hope, this is fine for you - otherwise please let me know.

I am also planning to add a security.md file just like you did for acaudit, but I refrain from creating a PR now to not further mess with your ways-of-working. 😀

tommystendahl:
Hi @gkunz 
Branching strategies are different in ecaudit and ecchronos, I think its enough if we do this only on master here.

tommystendahl (approved):

Issue #616, opened 2023-11-24
Build Ecchronos with Java 11


VictorCavichioli:
**Story Description**:

During the investigation of Ecchronos support for Java 17, it was observed that the product does not have a release with code compiled in Java 11. Therefore, to ensure there is at least one release version of Ecchronos in Java 11, the current code needs to be updated, compiled, and tested with Java 11.

Major changes may include updating versions of certain Maven packages, configuring the Maven-compiler-plugin, and any code changes that need to be identified during the development of this User Story. Following this, the following steps can be taken for validation:

1. Compile the project with Java 11.
2. Run unit tests.
3. Run integration tests.

**Acceptance Criteria**:

1. **Successful Compilation**: The project code must compile without errors using Java 11.
2. **Minimal Changes**: The provided code should include only what is necessary for the scope of the task; any changes outside the scope must be included in a separate issue.

**Definition of Done**:

The code must adhere to the defined quality standards, undergo a thorough review, and pass testing.

**Notes**:

_Not Applicable_

VictorCavichioli:
I will take this

PR #617, opened 2023-11-24
Build ecChronos With Java 17 - Issue #607

### LINKED ISSUE #607: Investigate Java 17
Investigate if ecChronos is working with java 17. If it doesn't then create issues for all work that needs to be done in order to be able to run with java 17.

VictorCavichioli:
In order to provide support for Java 17 in EcChronos, identify the minimal changes necessary for the project to compile and run with Java 17. Additionally, implement these changes and any necessary updates to Maven packages.

This change must be merged just after a release of ecChronos with the Issue #616 done.

Closes #607

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -39,8 +39,11 @@ jobs:
          - name: "Standalone integration 5.0-alpha1"
            test_suite: 'verify -P docker-integration-test,standalone-integration-tests -Dit.cassandra.version=5.0-alpha1 -DskipUTs'
            artifacts_dir: "standalone-integration/target"
+         - name: "Docker Integration"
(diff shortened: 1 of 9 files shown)


jwaeab (approved):

jwaeab (approved):

tommystendahl:
I don't think we should drop java11 support yet. It would be great if we could support both java 11 and java 17.

jwaeab:
Any decision on this one?

VictorCavichioli:
After some more discussions, we saw that we cannot we an alpha version, even if it's just for testing, so as the fix for jamm was done in version 5 we will wait for an stable release to keep working on this, related issue: https://issues.apache.org/jira/browse/CASSANDRA-16304

tommystendahl (approved):

PR #624, opened 2023-12-04
Build Ecchronos with Java 11 - Issue #616

### LINKED ISSUE #616: Build Ecchronos with Java 11

VictorCavichioli:
In order to provide support for Java 11 in EcChronos, identify the minimal changes necessary for the project to compile and run with Java 11. Additionally, implement these changes and any necessary updates to Maven packages.

Closes #616

### DIFF
--- a/connection/pom.xml
+++ b/connection/pom.xml
@@ -76,8 +76,8 @@
             <plugin>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
+                    <source>11</source>
+                    <target>11</target>
                 </configuration>
             </plugin>
         </plugins>
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Version 5.0.0 (Not yet released)
 
+* Build Ecchronos with Java 11 - Issue 616
 * Bump logback from 1.2.10 to 1.2.13 (CVE-2023-6378) - Issue #622
 * Progress for on demand repair jobs is either 0% or 100% - Issue #593
 * Make priority granularity configurable - Issue #599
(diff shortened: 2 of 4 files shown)


connection/pom.xml
```diff
@@ -76,8 +76,8 @@
             <plugin>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
+                    <source>11</source>
```
tommystendahl, line 79: Do we need to set source and target here, would we not inherit this setting from the parent pom.xml? If we do we should not set source and target in any pom-file except the parent pom. If not we should set it in all the pom-files, or?

VictorCavichioli: If we set the source and target configuration in the parent POM, the child modules will inherit these values by default. So yeah, in this case we do not need to repeat the configuration in the individual child module POMs unless if we want to override the configuration for a specific module, as we were doing previously with Java 8, so maybe we can remove all of this configuration in the child modules.

https://maven.apache.org/pom.html#POM_Relationships
https://maven.apache.org/pom.html#Inheritance

tommystendahl: Then I propose that we remove it from all the child modules and just set it in the parent pom.

PR #656, opened 2024-03-27
Implement RetryPolicy to Improve ecchronos Robustness - Issue #650

### LINKED ISSUE #650: [Robustness] Containerized ecchronos restarting tomcat when Cassandra peer is overloaded
Story Description:

1. Setup
* CIL Datacenter with 28 pods
* Traffic simulator producing enough traffic "on the verge"of overload, so that if one pod is down, the system starts to reject requests due to ECOP inbound

2. Step
* Provoke a complete restart of the datacenter with the following command:
kubectl get pod | grep rack | awk '{ print $1 }' | xargs kubectl delete pod

3. Expected Result
* I expect that the Tomcat server that is running ecchronos does not get killed, restarting the ecchronos container.
* I expect that while the pods are starting up and the cassandra container from the pods are starting up and getting overloaded, this does not make ecchronos to get killed.
* In the end, I expect that the pods are able to get up and not be locked at startup

Acceptance Criteria:

* Ecchronos does not stop/break when cassandra is overloaded
* The target DC doesn't get stuck in a complete restart

Definition of Done:

Repeating the same test case from the "Story Description" and the ecchronos containers do not restart during the attempts

Log Excerpt:

Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.ericsson.bss.cassandra.ecchronos.application.spring.RepairHistoryBean$$EnhancerBySpringCGLIB$$ed24782f]: Constructor threw exception; nested exception is com.datastax.oss.driver.api.core.DriverTimeoutException: Query timed out after PT2S

{"version":"1.0.0","timestamp":"2024-03-19T14:35:57.949Z","severity":"warning","service_id":"eric-bss-cil-database-cd","message":"[main] [org.springframework.beans.factory.support.DisposableBeanAdapter] Invocation of close method failed on bean with name 'defaultRepairConfigurationProvider': java.lang.NullPointerException"}
{"version":"1.0.0","timestamp":"2024-03-19T14:35:57.952Z","severity":"info","service_id":"eric-bss-cil-database-cd","message":"[main] [org.apache.catalina.core.StandardService] Stopping service [Tomcat]"}

VictorCavichioli:
Closes #650

### DIFF
--- a/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
+++ b/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
@@ -30,6 +30,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.ericsson.bss.cassandra.ecchronos.application.config.Config;
+import com.ericsson.bss.cassandra.ecchronos.application.config.RetryPolicy;
 import com.ericsson.bss.cassandra.ecchronos.application.config.security.Security;
 import com.ericsson.bss.cassandra.ecchronos.connection.NativeConnectionProvider;
 import com.ericsson.bss.cassandra.ecchronos.connection.impl.LocalNativeConnectionProvider;
@@ -81,7 +82,8 @@ public class DefaultNativeConnectionProvider implements NativeConnectionProvider
                 .withNodeStateListener(defaultRepairConfigurationProvider);
 
         myLocalNativeConnectionProvider = establishConnection(nativeConnectionBuilder,
-                host, port, nativeConfig.getTimeout().getConnectionTimeout(TimeUnit.MILLISECONDS));
+                host, port, nativeConfig.getTimeout().getConnectionTimeout(TimeUnit.MILLISECONDS),
+                nativeConfig.getRetryPolicy());
     }
 
     public DefaultNativeConnectionProvider(final Config config,
@@ -98,11 +100,10 @@ public class DefaultNativeConnectionProvider implements NativeConnectionProvider
             final LocalNativeConnectionProvider.Builder builder,
             final String host,
             final int port,
-            final long timeout)
+            final long timeout,
+            final RetryPolicy retryPolicy)
     {
-        long startTime = System.currentTimeMillis();
-        long endTime = startTime + timeout;
-        while (endTime > System.currentTimeMillis())
+        for (int count = 1; count <= retryPolicy.getMaxAttempts(); count++)
         {
             try
             {
@@ -110,19 +111,33 @@ public class DefaultNativeConnectionProvider implements NativeConnectionProvider
             }
             catch (AllNodesFailedException | IllegalStateException e)
             {
+                LOG.warn("Unable to connect through CQL using {}:{}, retrying", host, port);
+                long delay = retryPolicy.currentDelay(count);
+                long currentTime = System.currentTimeMillis();
+                long endTime = currentTime + timeout;
                 try
                 {
-                    LOG.warn("Unable to connect through CQL using {}:{}, retrying", host, port);
-                    if (timeout != 0)
+                    if (currentTime + delay > endTime)
                     {
-                        LOG.debug("Connection failed, retrying in 5 seconds", e);
+                        String msg = String.format(
+                                "Connection failed in attempt %d of %d. Retrying in 5 seconds.",
+                                count, retryPolicy.getMaxAttempts());
+                        LOG.warn(msg);
                         Thread.sleep(SLEEP_TIME);
+                    }
+                    else
+                    {
+                        String msg = String.format(
+                                "Connection failed in attempt %d of %d. Retrying in %d seconds.",
+                                count, retryPolicy.getMaxAttempts(), TimeUnit.MILLISECONDS.toSeconds(delay));
+                        LOG.warn(msg);
+                        Thread.sleep(delay);
                     }
                 }
                 catch (InterruptedException e1)
                 {
                     LOG.error("Unexpected interrupt while trying to connect to Cassandra", e1);
-                    throw e;
+                    throw new RuntimeException(e1);
                 }
             }
         }
--- a/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
+++ b/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
@@ -97,51 +97,62 @@ public class DefaultNativeConnectionProvider implements NativeConnectionProvider
     }
 
     private static LocalNativeConnectionProvider establishConnection(
-            final LocalNativeConnectionProvider.Builder builder,
-            final String host,
-            final int port,
-            final long timeout,
-            final RetryPolicy retryPolicy)
+        final LocalNativeConnectionProvider.Builder builder,
+        final String host,
+        final int port,
+        final long timeout,
+        final RetryPolicy retryPolicy)
     {
-        for (int count = 1; count <= retryPolicy.getMaxAttempts(); count++)
+        for (int attempt = 1; attempt <= retryPolicy.getMaxAttempts(); attempt++)
         {
             try
             {
-                return builder.build();
+                return tryEstablishConnection(builder);
             }
             catch (AllNodesFailedException | IllegalStateException e)
             {
-                LOG.warn("Unable to connect through CQL using {}:{}, retrying", host, port);
-                long delay = retryPolicy.currentDelay(count);
-                long currentTime = System.currentTimeMillis();
-                long endTime = currentTime + timeout;
-                try
-                {
-                    if (currentTime + delay > endTime)
-                    {
-                        String msg = String.format(
-                                "Connection failed in attempt %d of %d. Retrying in 5 seconds.",
-                                count, retryPolicy.getMaxAttempts());
-                        LOG.warn(msg);
-                        Thread.sleep(SLEEP_TIME);
-                    }
-                    else
-                    {
-                        String msg = String.format(
-                                "Connection failed in attempt %d of %d. Retrying in %d seconds.",
-                                count, retryPolicy.getMaxAttempts(), TimeUnit.MILLISECONDS.toSeconds(delay));
-                        LOG.warn(msg);
-                        Thread.sleep(delay);
-                    }
-                }
-                catch (InterruptedException e1)
-                {
-                    LOG.error("Unexpected interrupt while trying to connect to Cassandra", e1);
-                    throw new RuntimeException(e1);
-                }
+                handleRetry(attempt, retryPolicy, host, port);
             }
         }
-        return builder.build();
+        throw new RuntimeException("Failed to establish connection after all retry attempts.");
+    }
+
+    private static LocalNativeConnectionProvider tryEstablishConnection(
+        final LocalNativeConnectionProvider.Builder builder)
+    {
+        try
+        {
+            return builder.build();
+        }
+        catch (AllNodesFailedException | IllegalStateException e)
+        {
+            throw e;
+        }
+    }
+
+    private static void handleRetry(
+        final int attempt,
+        final RetryPolicy retryPolicy,
+        final String host,
+        final int port)
+    {
+        LOG.warn("Unable to connect through CQL using {}:{}, retrying.", host, port);
+        long delay = retryPolicy.currentDelay(attempt);
+        String msg = String.format(
+                "Connection failed in attempt %d of %d. Retrying in %d seconds.",
+                attempt, retryPolicy.getMaxAttempts(), TimeUnit.MILLISECONDS.toSeconds(delay));
+        LOG.warn(msg);
+        try
+        {
+            Thread.sleep(delay);
+        }
+        catch (InterruptedException e1)
+        {
+            LOG.error(
+                "InterruptedException caught during the delay time, while trying to reconnect to Cassandra. Reason: {}",
+                e1);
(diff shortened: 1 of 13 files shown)


application/src/main/resources/ecc.yml
```diff
@@ -32,8 +32,12 @@ connection:
     ## Connection is tried every five second until either the timeout time passes or the connection is successful.
     ##
     timeout:
-      time: 0
+      time: 60
       unit: seconds
+    retryPolicy:
```
jwaeab, line 37: Some kind of validation for the settings would probably be in place. As of now (considering the code), what would the setting maxDelay of 30 bring, when using the other defalut of 5/5?

jwaeab, line 37: Perhaps a unit setting for the retryPolicy settings too?

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
```diff
@@ -98,31 +100,44 @@ private static LocalNativeConnectionProvider establishConnection(
...
+                long endTime = currentTime + timeout;
                 try
                 {
-                    LOG.warn("Unable to connect through CQL using {}:{}, retrying", host, port);
-                    if (timeout != 0)
+                    if (currentTime + delay > endTime)
                     {
-                        LOG.debug("Connection failed, retrying in 5 seconds", e);
+                        String msg = String.format(
+                                "Connection failed in attempt %d of %d. Retrying in 5 seconds.",
```
jwaeab, line 123: Hardcoded 5. Replace with a %d instead and adding the sleep time, just in case the sleep time is changed later on.

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
```diff
@@ -98,31 +100,44 @@ private static LocalNativeConnectionProvider establishConnection(
...
             final String host,
             final int port,
-            final long timeout)
+            final long timeout,
+            final RetryPolicy retryPolicy)
     {
-        long startTime = System.currentTimeMillis();
-        long endTime = startTime + timeout;
-        while (endTime > System.currentTimeMillis())
+        for (int count = 1; count <= retryPolicy.getMaxAttempts(); count++)
```
jwaeab, line 106: Preference perhaps, but I would rename "count" to "attempt" just to be crystal clear of what it is. ;-)

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
```diff
@@ -98,31 +100,44 @@ private static LocalNativeConnectionProvider establishConnection(
...
+                        String msg = String.format(
+                                "Connection failed in attempt %d of %d. Retrying in %d seconds.",
+                                count, retryPolicy.getMaxAttempts(), TimeUnit.MILLISECONDS.toSeconds(delay));
+                        LOG.warn(msg);
+                        Thread.sleep(delay);
+                    }
                 }
                 catch (InterruptedException e1)
                 {
                     LOG.error("Unexpected interrupt while trying to connect to Cassandra", e1);
```
jwaeab, line 139: At this point, no reconnection is actually tried here, where the LOG.error is a bit misleading. The InterruptedException is thrown if the Thread.sleep is interrupted. See below for a suggestion.

jwaeab, line 139: Not exactly a part of this commit, but I would change the wording to something like "InterruptedException caught during the delay time, while trying to reconnect to Cassandra. Reason: {}", e1
"Unexpected interrupt" sound very vague.

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/DefaultNativeConnectionProvider.java
```diff
@@ -98,31 +100,44 @@ private static LocalNativeConnectionProvider establishConnection(
...
+            final RetryPolicy retryPolicy)
     {
-        long startTime = System.currentTimeMillis();
-        long endTime = startTime + timeout;
-        while (endTime > System.currentTimeMillis())
+        for (int count = 1; count <= retryPolicy.getMaxAttempts(); count++)
         {
             try
             {
                 return builder.build();
```
jwaeab, line 110: There are two return builder.build() in this function. Could refactoring would be in place? I mean, if the first one fails (even after retrying), wouldn't the risk of the last one failing as well be somewhat heightened?

jwaeab (changes requested):
- CHANGES.md needs to be updated.
- Docs should probably have something said about this.
- Why make a new separate section "("retryPolicy") for this, why not just append to the timeout section (with attempts etc)?

application/src/main/resources/ecc.yml
```diff
@@ -32,8 +32,12 @@ connection:
...
+      maxAttempts: 5
+      delay: 5
+      maxDelay: 30
```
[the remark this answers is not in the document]
VictorCavichioli: Yes you can, but considering an overloaded cluster, as the delay is calculated based on the attempt count, it can return a a really big number depending on the configuration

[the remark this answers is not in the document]
jwaeab: In the default case above, the practical max delay will be 25 (5*5), so 30 will never happen. If you set attempts/delay to something bigger than 5*5 - like 6*6 - you will already know that the max delay of 30 will be hit, so it is meaningless to set attempts/delay to anything bigger. And not to forget, you still have the overall timeout to consider.

I'm just trying to understand the actual benefit here for introducing the entire RetryPolicy class as it is. For instance, why don't you just use the current timeout setting/functionality as it is, and then "divide" that into dynamic delays instead of the hardcoded delay of 5 for each retry? So, you could do the first retry after 5 secs, the next after 10 secs and so on until the timeout has been spent.

Example, you could add like a retry_delta setting if you want the back off time to grow for each retry. So, a delta of 5 would give the scenario above (5, 10, 15, etc) and a setting of 0 would give the current behaviour (5, 5, 5, etc).

    timeout:
      time: 60
      retry_delta: 5
      unit: seconds

The DefaultNativeConnectionProvider constructor could probably be a bit refactored as well - in general- to be cleaner than it is now.

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/config/RetryPolicy.java
```diff
@@ -0,0 +1,88 @@
...
+        myDelay = TimeUnit.SECONDS.toMillis(delay);
+        myMaxDelay = TimeUnit.SECONDS.toMillis(maxDelay);
+    }
+
+    public RetryPolicy()
+    {
+
+    }
+
+    public final long currentDelay(final Integer count)
```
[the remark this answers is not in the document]
VictorCavichioli: Yes, the intention is to wait longer with each attempt, especially in scenarios where Cassandra might be overloaded to prevent ecChronos from breaking. However, we also set a maxDelay for this purpose, to prevent a single interval from having too high a number. This can be configured through the ecc.yml.

application/src/main/resources/ecc.yml
VictorCavichioli: It was done in the calculation of the intervals, if the calculated delay is greater than maxDelay, the maxDelay will be used instead.

VictorCavichioli: Done

VictorCavichioli:
> * CHANGES.md needs to be updated.
Done
> * Docs should probably have something said about this.
Done
> * Why make a new separate section "("retryPolicy") for this, why not just append to the timeout section (with attempts etc)?
Timeout now is used to calculate attempt delay, for sure we can do it together, but I think we are talking about different things, so two separate configurations, that's what I think :D

application/src/main/resources/ecc.yml
```diff
@@ -27,16 +27,23 @@ connection:
     host: localhost
     port: 9042
     ##
-    ## Connection Timeout for CQL.
+    ## Connection Timeout for an CQL attempt.
     ## Specify a time to wait for cassandra to come up.
-    ## Connection is tried every five second until either the timeout time passes or the connection is successful.
+    ## Connection is tried based on retry policy delay calculations. Each connection attempt will use timeout to calculate CQL Connection proccess delay.
```
jwaeab, line 32: "will use timeout" -> "will use the timeout"
Connection -> connection
proccess -> process

application/src/main/resources/ecc.yml
```diff
@@ -27,16 +27,23 @@ connection:
...
     ##
     timeout:
       time: 60
       unit: seconds
     retryPolicy:
+      ## Max number of attempts that ecChronos will try to connect with Cassandra.
```
jwaeab, line 38: remove "that"

application/src/main/resources/ecc.yml
```diff
@@ -27,16 +27,23 @@ connection:
...
       maxAttempts: 5
+      unit: seconds
+      ## Delay use to wait between an attempt and another, this value will be multiplied by the current attempt count.
+      ## If the calculated delay is greater than maxDelay, maxDelay will be used instead of the calculated delay
```
jwaeab, line 42: This is not correctly described, right? You do a power of 2 to the attempt, and then multiply the delay value to it. As in:
myDelay * Math.pow(2, attempt)

Missing a period at the end.

application/src/main/resources/ecc.yml
```diff
@@ -27,16 +27,23 @@ connection:
...
-    ## Connection is tried every five second until either the timeout time passes or the connection is successful.
+    ## Connection is tried based on retry policy delay calculations. Each connection attempt will use timeout to calculate CQL Connection proccess delay.
     ##
     timeout:
       time: 60
       unit: seconds
     retryPolicy:
+      ## Max number of attempts that ecChronos will try to connect with Cassandra.
       maxAttempts: 5
+      unit: seconds
```
jwaeab, line 40: I would move this setting below maxDelay instead, since it is how the style is in the rest of the yml file. It is more logical to have after the settings it is related to.

docs/RETRY_POLICY.md
```diff
@@ -0,0 +1,29 @@
+# Retry Policy
+
+The Retry Policy configures the connection retry behavior of ecChronos with Cassandra. The Timeout specifies the period of time ecChronos will wait for Cassandra to start per attempt. The retry policy sets the maximum number of connection attempts (maxAttempts), which by default is 5, specifying the time unit as seconds by deafult. Additionally, it determines the delay between each connection attempt (delay), multiplied by the current attempt count.
```
jwaeab, line 3: The power of 2 thing here as well

docs/RETRY_POLICY.md
```diff
@@ -0,0 +1,29 @@
...
+
+If the calculated delay exceeds the allowed maximum (maxDelay), the maximum will be used. The maxDelay limits the maximum time a connection attempt can wait to execute. If set to 0, it disables maxDelay, and the delay interval will be calculated based on the number of attempts and the default delay. These settings aim to optimize the efficiency of connection attempts, ensuring the system waits an appropriate amount of time and makes a reasonable number of attempts before giving up.
+
```
jwaeab, line 6: Perhaps you could write an example here of a "timeline" of the connection attempts to visualize how the settings actually affects the delays/timeout? You know, if you pretend that all connection attempts fails. I think that would be good, because it easily gets just a lot of number juggling otherwise. ;-)

docs/RETRY_POLICY.md
```diff
@@ -0,0 +1,29 @@
...
+## Example on ecc.yml
+
+```yaml
+    ##
+    ## Connection Timeout for an CQL attempt.
```
jwaeab, line 11: "a CQL"

docs/RETRY_POLICY.md
```diff
@@ -0,0 +1,29 @@
...
+    ## Specify a time to wait for cassandra to come up.
+    ## Connection is tried based on retry policy delay calculations. Each connection attempt will use timeout to calculate CQL Connection proccess delay.
```
jwaeab, line 13: Same as above - in the ecc.yml

docs/RETRY_POLICY.md
```diff
@@ -0,0 +1,29 @@
...
+    ##
+    timeout:
+      time: 6
+      unit: minutes
+    retryPolicy:
+      ## Max number of attempts that ecChronos will try to connect with Cassandra.
+      maxAttempts: 10
+      unit: minutes
```
jwaeab, line 21: Move to after maxDelay

docs/RETRY_POLICY.md
```diff
@@ -0,0 +1,29 @@
...
+      ## Delay use to wait between an attempt and another, this value will be multiplied by the current attempt count.
```
jwaeab, line 22: Same as above - in the ecc.yml

application/src/main/resources/ecc.yml
```diff
@@ -27,16 +27,23 @@ connection:
...
+      ## Delay use to wait between an attempt and another, this value will be multiplied by the current attempt count.
+      ## If the calculated delay is greater than maxDelay, maxDelay will be used instead of the calculated delay
       delay: 5
+      ## The max delay that one attempt can wait until run.
```
jwaeab, line 44: "Maximum delay before the next connection attempt is made."

jwaeab (changes requested):
Add a test case where you actually test each delay value for a number of attempts, where the max delay is also hit.

jwaeab (approved):
LGTM!

tommystendahl (approved):

Issue #533, opened 2023-08-08
Investigate debug logs of ecchronos


masokol:
Investigate debug logs of ecchronos to see how much is actually logged. Maybe some stuff should be throttled or removed/moved to trace.

masokol:
There's a lot of irrelevant logs on debug lvl. We should move them to trace instead.

VictorCavichioli:
I will take this

DanielwEriksson:
This is the suggestion

epkdaek@elx721027t9:~/github/ecchronos$ git grep "LOG.debug"
INFO: application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/ECChronosInternals.java:            LOG.debug("Table {} last repaired at {}", tableReference, lastRepairedAt);
INFO: application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/ECChronosInternals.java:            LOG.debug("Table {} remaining repair time {}", tableReference, remainingRepairTime);
INFO: application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/config/ConfigRefresher.java:        LOG.debug("Watching for changes in {}", absoluteFilePath);
INFO: application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/config/ConfigRefresher.java:                LOG.debug("Watch service has been closed");
DEBUG: application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/config/ConfigRefresher.java:        LOG.debug("Received event for {}/{}", baseDirectory, file);
DEBUG: application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/spring/CassandraHealthIndicator.java:            LOG.debug("Unable to connect over JMX", e);
DEBUG: application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/spring/CassandraHealthIndicator.java:            LOG.debug("Unable to connect over CQL", e);
INFO: connection.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/connection/impl/LocalJmxConnectionProvider.java:        LOG.debug("Connecting JMX through {}, credentials: {}, tls: {}", jmxUrl, authEnabled, tlsEnabled);
INFO: connection.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/connection/impl/LocalJmxConnectionProvider.java:        LOG.debug("Connected JMX for {}", jmxUrl);
[...]

PR #660, opened 2024-05-07
Investigate debug logs of ecchronos - Issue #533

DanielwEriksson:


VictorCavichioli (changes requested):
I think maybe you need to add on logback.xml the TRACE level, I think that if you're not doing it the logs will be just ignored.

Please, test it.

jwaeab (commented):
A couple of generic questions before dwelling into the actual code review:

- What kind of criteria(s) have you based these changes on?
- Did you check with Mackan what he meant by "throttling"? I see no code/comments relating to this. There is a ThrottlingLogger class in ecChronos, perhaps he meant make use of that more?
- While doing this, have you also checked if any of the other log levels need to be changed in any way (warn, info, error etc)? Perhaps you have traces that actually should be debug etc. But this kind of relates to my first point.
- In a quick read through of the code, I find odd LOGs in general.

DanielwEriksson:
To quote macan when I asked for a Definition of Done for this task.
"DoD är, flytta alla debug loggar som inte hjälper med troubleshooting till trace."
This is just a first suggestion otherwise we have to discuss each debug entry in the log. I have tried to keep those that I think is part of a flow. Those that I think is just some random information I have moved to trace.

jwaeab:
Debug is perhaps more used for isolated events, so those can be debugged more easily. Trace on the other hand is perhaps mostly used when trace execution/flows/request-response-trails/etc.
If something is regarded "random information", then it is probably not useful for either debugging or tracing (or logging to begin with).

jwaeab:
Part of this will be done in issue #666

Issue #661, opened 2024-05-14
Spring Framework URL Parsing with Host Validation is vulnerable


dapc11:
Hello,

With the current dependency tree, of EcChronos 5.0.2, there are three vulnerabilities reported on Spring Framework.
I've looked into the code, but there is no usage of the vulnerable class. However, it would be nice to uplift the vulnerable component to not clutter the reports of security scanners and cause unnecessary administration work downstream.

- https://spring.io/security/cve-2024-22262
- https://spring.io/security/cve-2024-22259
- https://spring.io/security/cve-2024-22243

PR #663, opened 2024-05-15
Override Spring-Web dependency to fix vulnerabilities

VictorCavichioli:
Issue - #661

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -97,7 +97,7 @@
         <io.prometheus.simpleclient>0.16.0</io.prometheus.simpleclient>
         <jakarta.validation-api.version>2.0.2</jakarta.validation-api.version>
         <swagger-annotations.version>2.2.11</swagger-annotations.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -97,6 +97,7 @@
         <io.prometheus.simpleclient>0.16.0</io.prometheus.simpleclient>
         <jakarta.validation-api.version>2.0.2</jakarta.validation-api.version>
         <swagger-annotations.version>2.2.11</swagger-annotations.version>
+        <org.springframework.web.version>5.3.33</org.springframework.web.version>
```
jwaeab, line 100: Why not 5.3.34? Isn't that the latest?

VictorCavichioli: Yes, I chose 5.3.33 because it's the patched version for the vulnerability, but I can update to the latest version.

jwaeab (approved):

PR #670, opened 2024-06-04
Issue666 - Define logging levels formally 

DanielwEriksson:


docs/CONTRIBUTING.md
```diff
@@ -27,6 +27,34 @@ Formatting rules for eclipse can be found [here](../code_style.xml).
 Provided patches should be contained and should not modify code outside of the scope of the patch.
 This will make it quicker to perform reviews and merging the pull requests.
 
+## Logging
+
+Use this as a guideline for what to log for each level, also use the isXyzEnabled (isTraceEnabled, isDebugEnabled etc)
+before logging if the log message that contains anything else than already defined variables.
```
jwaeab, line 33: I would put the "isEnabled" text below the table, right before the actual example.

docs/CONTRIBUTING.md
```diff
@@ -27,6 +27,34 @@ Formatting rules for eclipse can be found [here](../code_style.xml).
...
+| error     | A service fail in the sense it cannot response to requests or data processed cannot be trusted/used.  <br>Example: Connection attempts that ultimate fail. Crucial resources not available.                                                                                                                                                                                                                                 |
+| off       | No levels will be logged <br> Example: -                                                                                                                                                                                                                                                                                                                    |                                                                                                                                                                                                                                                |
+
+This is a code example where to use isEnabled function
+ ```
+if (LOG.isDebugEnabled())
```
jwaeab, line 47: Format the code example better. It is hard to read when everything is left aligned.

ecchronos-binary/src/resources/logback.xml
```diff
@@ -65,6 +69,5 @@
     <root level="DEBUG">
         <appender-ref ref="STDOUT" />
         <appender-ref ref="FILE_INFO" />
-        <appender-ref ref="FILE_DEBUG" />
```
jwaeab, line 68: Since this debug is removed from default, I would also write that in the documentation (where appropriate). That is, clarify that the info level is the only log level enabled by default and if debug is necessary, describe how to enable it.

DanielwEriksson (commented):
looks good to me but that is my opinion

DanielwEriksson (commented):
Looks good to me but that is just my opinion

docs/CONTRIBUTING.md
```diff
@@ -27,6 +27,41 @@ Formatting rules for eclipse can be found [here](../code_style.xml).
...
+      if (old != null)
+      {
+         next -= old.getEstimatedRepairTime();
+      }
+      LOG.debug("Table {} partially repaired at {}, next repair at/after {}", myTableReference,
+     <b>MY_DATE_FORMAT.get().format(new Date(repairedAt))</b>, <b>MY_DATE_FORMAT.get().format(new Date(next))</b>);
+}
+</pre>
+<pre>
+if (LOG.isInfoEnabled())
```
jwaeab, line 58: Bad example! An info log should always be printed (and be printable) in production code.

docs/CONTRIBUTING.md
jwaeab: You still haven't "resolved" this. Why complicate this example?

jwaeab (changes requested):
Instead of asking on each line, I will make this generic:
1) You have changed a lot of detailed traces to debug. Why?
2) The same goes from debug to info. Why?
3) Did you do some testing on this to check the outoput? How did you do that?

DanielwEriksson (commented):
looks good to me but that is my opinion, I have made changes to align with the guidelines

PR #703, opened 2024-08-26
Deprecate cassandra-all to use testContainers

### LINKED ISSUE #701: Deprecate cassandra-all to use testContainers instead
Story Description:
We are facing issues to upgrade ecchronos sidecar to java 17 because of some dependencies of cassandra-all, after issue #682 in agent poc we saw that we can use it on unit-tests to avoid these problems, the goal is to deprecate cassandra-all to update the sidecar to java 17.

Acceptance Criteria:

NA

Definition of Done:

Tests passing, code approved by maintainers

Notes:

This class can be used as reference: https://github.com/Ericsson/ecchronos/blob/agent/master/data/src/test/java/com/ericsson/bss/cassandra/ecchronos/data/utils/AbstractCassandraTest.java

VictorCavichioli:
Closes #701


VictorCavichioli:
I was thinking about upgrading the jackson and snakeyaml versions together with this PR but because of some others dependencies the change breaks the openapi, so I removed it, as I don't have time now to look at this update and this is not a priority I created this new issue so me or someone else can look into that in some moment https://github.com/Ericsson/ecchronos/issues/704

tommystendahl (approved):

Issue #678, opened 2024-06-17
Generate Unique EcChronos ID


VictorCavichioli:
Story Description:
As the ecchronos_id is the partition key for nodes_sync table, the goal is to find a way to create a unique identification for the instance, so when a node is acquired by that instance the nodes_sync must receive a static value, it must have the same id even if the instance goes down.

Acceptance Criteria:

- [ ] Code designed as described
- [ ] Test cases
- [ ] Documentation updated

Definition of Done:

Code designed as described and approved by maintainers

Notes:

Related to #652 and #672

PR #713, opened 2024-09-19
Generate Unique EcChronos ID #678

SajidRiaz138:
Instance Name can be configure in ecc.yml file as partiton key and each instance must have unique name.

### DIFF
--- a/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/config/connection/AgentConnectionConfig.java
+++ b/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/config/connection/AgentConnectionConfig.java
@@ -37,6 +37,7 @@ public final class AgentConnectionConfig
     private RackAware myRackAware = new RackAware();
     private HostAware myHostAware = new HostAware();
     private Class<? extends DefaultLoadBalancingPolicy> myDatacenterAwarePolicy = DataCenterAwarePolicy.class;
+    private String myInstanceName;
 
     /**
      * Default constructor for AgentConnectionConfig.
@@ -44,6 +45,40 @@ public final class AgentConnectionConfig
     public AgentConnectionConfig()
     {
 
+    }
+
+    /**
+     * Gets unique ecchronos instance name.
+     *
+     * @return the instance name.
+     */
+    @JsonProperty("instanceName")
+    public String getInstanceName()
+    {
+        return myInstanceName;
+    }
+
+    /**
+     * Sets unique instance name.
+     *
+     * @param  instanceName
+     *         the instance name as a string.
+     * @throws ConfigurationException
+     *         if the provided instance name is invalid.
+     */
+    @JsonProperty("instanceName")
+    public void setInstanceName(final String instanceName) throws ConfigurationException
+    {
+        if (instanceName == null || instanceName.isBlank())
+        {
+            throw new ConfigurationException(
+                    "Invalid instance name: "
+                            +
+                            instanceName
+                            +
+                            "\nInstance name must not be null or empty");
+        }
+        myInstanceName = instanceName;
     }
 
     /**
--- a/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/spring/BeanConfigurator.java
+++ b/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/spring/BeanConfigurator.java
@@ -96,7 +96,7 @@ public class BeanConfigurator
         Security security = getSecurityConfig();
         cqlSecurity.set(security.getCqlSecurity());
         jmxSecurity.set(security.getJmxSecurity());
-        ecChronosID = ECCHORONS_ID_PRE_STRING.concat(InetAddress.getLocalHost().getHostName());
+        ecChronosID = getConfiguration().getConnectionConfig().getCqlConnection().getAgentConnectionConfig().getInstanceName();
     }
 
     /**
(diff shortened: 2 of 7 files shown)


application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/spring/BeanConfigurator.java
```diff
@@ -64,7 +64,6 @@ public class BeanConfigurator
     private final AtomicReference<Security.CqlSecurity> cqlSecurity = new AtomicReference<>();
     private final AtomicReference<Security.JmxSecurity> jmxSecurity = new AtomicReference<>();
     private final ConfigRefresher configRefresher;
-    private final String ecChronosID;
```
VictorCavichioli, line 67: Idk about removing it, because we need to generate this ans pass for EccNodeMetadata if the instance is new, I think here we need to have some validation using the EccNodeMetadata, generate the unique id, and them passing for EccNodesSync

data/src/main/java/com/ericsson/bss/cassandra/ecchronos/data/sync/EccNodeMetadata.java
```diff
@@ -0,0 +1,134 @@
...
+import java.net.UnknownHostException;
+import java.util.List;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static com.datastax.oss.driver.api.querybuilder.QueryBuilder.bindMarker;
+import static com.datastax.oss.driver.api.querybuilder.QueryBuilder.selectFrom;
+
+public final class EccNodeMetadata
```
VictorCavichioli, line 35: EccNodesMetadata

VictorCavichioli (changes requested):
Missing updates on docs

data/src/main/java/com/ericsson/bss/cassandra/ecchronos/data/sync/EccNodeMetadata.java
```diff
@@ -0,0 +1,134 @@
...
+    private final PreparedStatement mySelectStatement;
+    private final PreparedStatement myCreateStatement;
+    private final PreparedStatement myTableUpdateStatement;
+    private final CqlSession mySession;
+    private final Node myNode;
+
+    public EccNodeMetadata(final List<Node> nodes, final CqlSession session)
+    {
+        mySession = session;
+        myNode = nodes.stream().findAny().get();
```
VictorCavichioli, line 54: What happens if the topology changes and you get a node that was not on the first configuration for a instance that was working?

let's say that in the configuration we are using datacenterAware, so we have a datacenter list of [d1, dc2] but now the topology changed and we added dc3 on the list, but it was not on the previously topology for this instance, so if this line picks up a node from dc3 it will generate a new ecchronos id for a instance that already has an ID

VictorCavichioli (changes requested):

tommystendahl (approved):

application/src/main/resources/ecc.yml
```diff
@@ -24,6 +24,8 @@ connection:
     ## to connect to. The application will use the configurations
     ## specified below, connecting to the listed hosts;
     agent:
+      ## Each ecchronos instance must have unique name
```
VictorCavichioli, line 27: I think it need to be more descriptive, also, explain in the architecture.md what is that and why we need it

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/config/connection/AgentConnectionConfig.java
```diff
@@ -37,6 +37,7 @@ public final class AgentConnectionConfig
     private RackAware myRackAware = new RackAware();
     private HostAware myHostAware = new HostAware();
     private Class<? extends DefaultLoadBalancingPolicy> myDatacenterAwarePolicy = DataCenterAwarePolicy.class;
+    private String myInstanceName;
```
VictorCavichioli, line 40: Should it have a default value? Like using the hostname/hostid

VictorCavichioli (changes requested):

application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/config/connection/AgentConnectionConfig.java
SajidRiaz138: I have made it mandatory and it would throw configuration error if someone will not set it. if we use default value it will defeat that purpose. so it is mandatory property and must be set.

docs/ARCHITECTURE.md
```diff
@@ -49,6 +49,7 @@ Configuration is available on ecc.yml in the below format.
 connection:
   cql:
     agent:
+      instanceName: Each instance must have unique instance name. This will be used as ecchronos_id (partition key in nodes_sync table)
```
VictorCavichioli, line 52: The field description should be with the text that explain the nodes sync table I guess, something like:

"To keep track about nodes and instances, the Agent implementation uses the table nodes_sync, that declare nodes by specific ecChronos instances, each instance must have unique instance name, that must be defined in `connection.cql.agent.instanceName` in ecc.yml. This will be used as ecchronos_id (partition key in nodes_sync table), to use the agent, is mandatory to create the table below:"

VictorCavichioli (changes requested):

PR #786, opened 2024-12-09
Create dependabot.yml

tommystendahl:


jwaeab (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
...
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "maven" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "github-actions" # See documentation for possible values
+    directory: "/" # Location of package manifests
```
yurnov, line 13: probably here it's enough to have `/.github/workflows`

tommystendahl: Actually, github documentation say that you should set it to / for github-actions despite the default location is /.github/workflows. My guess is that they simply hard coded the path and that this config doesn't change anything for github-actions.

yurnov: > Actually, github documentation say that you should set it to / for github-actions despite the default location is /.github/workflows. My guess is that they simply hard coded the path and that this config doesn't change anything for github-actions.

I have used that way in [my own project](https://github.com/yurnov/payment-qr-generator/blob/main/.github/dependabot.yml#L9) and it's works fine.

PR #858, opened 2025-01-29
Fix python version in test to 3.12

tommystendahl:
The latest version (3.29.2) of the python driver does not support python 3.13, once a new version of the python-driver supporting python 3.13 is available this change should be reverted.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -92,7 +92,7 @@ jobs:
       - name: Set up Python 3
         uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
         with:
-          python-version: 3.12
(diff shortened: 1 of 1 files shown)


.github/workflows/actions.yml
```diff
@@ -92,7 +92,7 @@ jobs:
       - name: Set up Python 3
         uses: actions/setup-python@0b93645e9fea7318ecaed2b359559ac225c90a2b # v5.3.0
         with:
-          python-version: 3
+          python-version: 3.12
```
jwaeab, line 95: Perhaps just adding some sort of comment here to why we explicitly set 3.12 here?

tommystendahl: done

jwaeab (approved):

Issue #1002, opened 2025-06-18
Enabled cipher suites are set twice


tommystendahl:
We set enabled cipher suites twice, at these places:
https://github.com/Ericsson/ecchronos/blob/893bfd040a47152f2d8bd6d916747b8ddc4d3bbd/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/ReloadingCertificateHandler.java#L110
https://github.com/Ericsson/ecchronos/blob/893bfd040a47152f2d8bd6d916747b8ddc4d3bbd/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/ReloadingCertificateHandler.java#L306
The first use javax and the second use the netty API, it would be nice if we could remove the first and keep the one using Netty API.

paulchandler:
Following discussions with Johan, we have decided to put this on hold.

This section is used for the Unit Tests, but there are no tests for the ciphers. So we are unsure whether it can be removed.

tommystendahl:
Since we use Netty I think we should stick to the Netty API as much as possible. I think the line using javax is a leftover that should have been removed when the switch to Netty was done. I have not investigated the test coverage in this area but before we start changing anything we should make sure that there are tests covering this.

PR #1008, opened 2025-06-24
#1002 Enabled cipher suites are set twice. Line removed

paulchandler:
Remove line setting cipher suite for the second time


paulchandler:
Currently it has been agreed not to continue with this request

Issue #921, opened 2025-03-25
Add CRL support


jwaeab:
Refer to this closed issue (https://github.com/Ericsson/ecchronos/issues/880)

VictorCavichioli:
Related with #652

paulchandler:
Completed with the previous PR

PR #1014, opened 2025-07-10
#921 Add crl support for certificates

paulchandler:
Add crl support for certificates, ported from the master branch.


jwaeab (commented):
Looks ok, but I can have another look once you have updated the branch with the latest stuff as well.

jwaeab (approved):

PR #1032, opened 2025-07-23
REST interface only shows the schedules for the last table added

paulchandler:
Fixes RepairSchedulerImpl so that it does not lose earlier table definitions


VictorCavichioli (approved):

VictorCavichioli (changes requested):
Add fix in CHANGES.md

VictorCavichioli (approved):

PR #1048, opened 2025-08-08
Agent/1045 Add behave tests

paulchandler:
This adds some behave tests for the new state REST interface 

Some of the common files have extra changes in that are part of the port of the behave changes to the new version.


paulchandler:
This is going to the wrong branch

PR #1049, opened 2025-08-08
Agent/1045

paulchandler:
This adds some behave tests for the new state REST interface

Some of the common files have extra changes in that are part of the port of the behave changes to the new version.

### DIFF
--- a/ecchronos-binary/src/test/behave/features/steps/ecc_rest_state.py
+++ b/ecchronos-binary/src/test/behave/features/steps/ecc_rest_state.py
@@ -29,10 +29,4 @@ from ecc_step_library.common import (
 def step_check_node_count(context, count):
     assert context.response is not None
     context.json = context.response.json()
-    print("Number of JSON objects:", len(context.json) )
-
-#    for obj in context.json:
-#        if obj["keyspace"] == keyspace and obj["table"] == table:
-#            context.id = obj["nodeID"]
-#            break
     assert len(context.json) == int(count)
--- a/ecchronos-binary/src/test/behave/features/steps/ecc_rest_state.py
+++ b/ecchronos-binary/src/test/behave/features/steps/ecc_rest_state.py
@@ -1,0 +1,38 @@
+#
+# Copyright 2025 Telefonaktiebolaget LM Ericsson
+#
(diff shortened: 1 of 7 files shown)


ecchronos-binary/src/test/behave/features/steps/ecc_rest_state.py
```diff
@@ -0,0 +1,38 @@
...
+@then("the number of nodes is {count}")
+def step_check_node_count(context, count):
+    assert context.response is not None
+    context.json = context.response.json()
+    print("Number of JSON objects:", len(context.json) )
+
+#    for obj in context.json:
+#        if obj["keyspace"] == keyspace and obj["table"] == table:
+#            context.id = obj["nodeID"]
+#            break
```
VictorCavichioli, line 37: Remove

ecchronos-binary/src/test/behave/features/steps/ecc_spring.py
```diff
@@ -0,0 +1,23 @@
+#
+# Copyright 2019 Telefonaktiebolaget LM Ericsson
```
VictorCavichioli, line 2: 2025

VictorCavichioli (changes requested):
As behave tests works now, please, add the action rule like in https://github.com/Ericsson/ecchronos/blob/ecchronos-6.0/.github/workflows/actions.yml#L45 so we know that all tests are working as expected

paulchandler:
> As behave tests works now, please, add the action rule like in https://github.com/Ericsson/ecchronos/blob/ecchronos-6.0/.github/workflows/actions.yml#L45 so we know that all tests are working as expected

Currently this requires a cluster to be setup manually, so it is not ready for this yet.

ecchronos-binary/src/test/behave/features/steps/ecc_rest_state.py
paulchandler: Done

ecchronos-binary/src/test/behave/features/steps/ecc_spring.py
paulchandler: Done

Issue #1052, opened 2025-08-12
Remove deprecated class LongToken


tommystendahl:
The LongToken class was deprecated in 1.0 and is not used anywhere.

BrianAcosta110325:
Could you please share a bit more context about this issue?
I’d like to understand where in the code or module it is used, and what you would recommend doing to address it.
I'm interested in solving it.

tommystendahl:
Hi @BrianAcosta110325 and welcome to the project.
The issue is about this class [LongToken](https://github.com/Ericsson/ecchronos/blob/ecchronos-6.0/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/utils/LongToken.java) and the corresponding test class. It was deprecated in this PR: [Reduce memory footprint](https://github.com/Ericsson/ecchronos/pull/55) six years ago for ecchronos 1.0 and scheduled for removal in 2.0 but it looks like that didn't happen.
I could not find any use of this class so I think it can just be removed, and then its just to verify that the project still builds. If there are any issues after removing the class and test class I think they should be simple to fix but we can analyze that if it happens.
Do you want me to assign the issue to you?

BrianAcosta110325:
Yes, please assign me the issue. I’ll start by locating and removing the LongToken class and its corresponding test class, then ensure the project builds successfully. If any related issues come up, I’ll analyze and address them accordingly.

BrianAcosta110325:
I’ve removed the LongToken class and its corresponding test files. After this change, the build passes for most modules.

LongTokenRange appears to still be in use, so I haven’t removed it—please confirm if it should remain.

Also, the Binary module test was already failing before the removal of LongToken, so this change did not cause that failure. Could you provide guidance on whether I should leave it as is?

tommystendahl:
Its merged now and everything looks good.

@BrianAcosta110325 Thank you very much for your contribution, nice work.

PR #1058, opened 2025-08-13
remove LongToken class and associated test file

BrianAcosta110325:
This PR addresses issue #1052 

The LongToken class and its corresponding tests have been removed. The build has been verified using mvn, and all modules pass except for the Binary module, which was already failing prior to this change.

LongTokenRange remains in the code, as it is still in use—please confirm if it should stay.


tommystendahl:
LongTokenRange is used, only LongToken should be removed.

I have triggered the tests on GitHub Actions, lets see what the result is there.

PR #1161, opened 2025-10-01
Separate Lock Failure Cache by Node to Prevent Cross-Node Lock Blocking

### LINKED ISSUE #1159: Separate Lock Failure Cache by Node to Prevent Cross-Node Lock Blocking
Story Description:

Currently, the LockCache implementation caches lock failures using a LockKey that only considers the datacenter and resource. This approach causes lock failures from one node to prevent lock acquisition attempts on other nodes within the same datacenter, leading to unnecessary job blocking and reduced repair throughput.

When a lock acquisition fails on one node, the failure is cached and prevents subsequent lock attempts for the same resource across all nodes in the datacenter, even though other nodes might be able to successfully acquire the lock. This behavior was observed during testing where jobs would get permanently stuck in "Trying to acquire lock" state due to cached failures from different nodes.

The enhancement should modify the lock failure caching mechanism to be node-specific, ensuring that a lock failure on one node does not prevent lock acquisition attempts on other nodes. This change should be implemented in the LockCache class and related lock management components.

Acceptance Criteria:

- The LockKey record should include node identification (e.g., node UUID or host ID) in addition to datacenter and resource to create node-specific cache entries.
- Lock failures cached for one node should not prevent lock acquisition attempts on different nodes within the same datacenter and resource combination.
- The cache expiry mechanism should continue to work as expected, clearing node-specific failures after the configured timeout period.
- Successful lock acquisition on any node should clear the cache entry only for that specific node, not affecting cached failures for other nodes.
- The enhancement should maintain backward compatibility with existing lock factory implementations (CASLockFactory, RepairLockFactoryImpl) without requiring changes to their interfaces.

Definition of Done:

- Code changes implemented in LockCache.java and related classes
- Unit tests added to verify node-specific caching behavior

Notes:

Related with #652

VictorCavichioli:
Closes #1159

### DIFF
--- a/core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/locks/CASLockFactory.java
+++ b/core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/locks/CASLockFactory.java
@@ -83,7 +83,7 @@ public final class CASLockFactory implements LockFactory, Closeable
     private final CASLockProperties myCasLockProperties;
     private final CASLockStatement myCasLockStatement;
     private final DistributedNativeConnectionProvider myNativeConnectionProvider;
-    private final long cacheExpiryTimeInSeconds;
+    private final long myCacheExpiryTimeInSeconds;
 
     CASLockFactory(final CASLockFactoryBuilder builder)
     {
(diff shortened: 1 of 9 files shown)


core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/locks/CASLockFactory.java
```diff
@@ -80,6 +82,8 @@ public final class CASLockFactory implements LockFactory, Closeable
 
     private final CASLockProperties myCasLockProperties;
     private final CASLockStatement myCasLockStatement;
+    private final DistributedNativeConnectionProvider myNativeConnectionProvider;
+    private final long cacheExpiryTimeInSeconds;
```
jwaeab, line 86: my...

;-)

PR #1211, opened 2025-10-31
Feature: Add Reload Topology Scheduler

VictorCavichioli:
This PR implements a comprehensive solution for handling dynamic topology changes in Cassandra clusters, ensuring that ecChronos can automatically detect and respond to node additions, removals, and status changes without requiring manual intervention or restarts.

The periodic reload mechanism solves a specific race condition:

1. Initial Startup: When ecChronos starts, it builds an initial topology map from currently available nodes
2. Node Initialization Window: If a Cassandra node is still initializing (not yet "UP"), it won't be included in this initial map
3. Driver Behavior: The Cassandra driver only sends onAdd events for nodes that weren't in the original topology
4. Missing Detection: Nodes that come online after being in the initial topology but were "DOWN" will trigger onUp events, but nodes that were never in the initial map won't trigger onAdd
5. Reload Solution: The periodic reload rebuilds the entire topology from scratch, ensuring no nodes are missed regardless of timing

### DIFF
--- a/ecchronos-binary/src/pylib/ecchronoslib/types.py
+++ b/ecchronos-binary/src/pylib/ecchronoslib/types.py
@@ -164,6 +164,7 @@ class RepairStats(object):
 
 
 class NodeSyncState(object):
+    # pylint: disable=too-few-public-methods
     def __init__(self, data):
         self.ecchronos_id = data["ecchronosId"] if "ecchronosId" in data else "<UNKNOWN>"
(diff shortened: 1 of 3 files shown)


jwaeab (commented):
Massive work - as usual, Victor! I think you just need to sort out the failing tests before we have a deeper look at it. ;-)

jwaeab (approved):

PR #1220, opened 2025-11-04
 #1219 Fix metrics endpoint returning empty response

### LINKED ISSUE #1219: Metrics endpoint returns empty response
Description:

The /metrics REST endpoint returns no metrics when called without the includedMetrics parameter, causing monitoring systems to receive empty responses.

1- Does the problem persist?
yes

2- Describe how to reproduce the issue:

-Enable HTTP ECC yaml statistics:
	statistics.enabled: True
	statistics.reporting.http.enabled: True
-Configure application yaml
	metricsServer.enabled: True
	metricsServer.port: 9501 #your port
-Call the /metrics endpoint without any includedMetrics parameter: GET /metrics
-Observe that the response body is empty instead of returning all available metrics

Detailed description:

• What is happening?
The /metrics endpoint returns an empty response when no includedMetrics parameter is provided.

• What did you expect to happen?
When calling /metrics without specifying includedMetrics, all available metrics should be returned, similar to standard Prometheus endpoints.

• What have you tried?
I installed the older version 5.0.4, and it worked; 6.0.0 has the issue.

• What version of ecChronos are you using?
6.0.0

• Was the problem detected during an upgrade or downgrade procedure?
No

• Have you checked the ecChronos documents?
Yes

• What do YOU think is the issue?
The issue is in MetricsREST.java where the code doesn't handle the case when includedMetrics parameter is null or empty. The Prometheus registry has different method signatures for scraping all metrics vs. specific metrics.
This probably was caused by the dependnecy update made on 6.0.0 version that changed the packaged used in this class.

Fix:
java
String scrapeResult = (includedMetrics == null || includedMetrics.isEmpty())
        ? myPrometheusMeterRegistry.scrape(contentType)
        : myPrometheusMeterRegistry.scrape(contentType, includedMetrics);

return ResponseEntityok()
        .header(CONTENT_TYPE, contentType)
        .body(scrapeResult);

cezarpaulo16:
## Description
Fixes issue where the `/metrics` REST endpoint returns empty response when called without the `includedMetrics` parameter or with an empty set.

## Problem
The current implementation always calls `myPrometheusMeterRegistry.scrape(contentType, includedMetrics)` even when `includedMetrics` is null or empty, which returns no metrics instead of all available metrics.

## Solution
- Check if `includedMetrics` is null or empty
- Call `scrape(contentType)` when no specific metrics are requested (returns all metrics)
- Call `scrape(contentType, includedMetrics)` when specific metrics are requested (returns filtered metrics)

## Testing
- Added comprehensive tests using real `PrometheusMeterRegistry` instead of mocks
- Tests cover null parameter, empty set, and filtering scenarios
- Refactored existing tests to eliminate duplication and improve maintainability

Closes #1219

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Version 6.1.4
 
+* Fix metrics endpoint returning empty response - Issue #1219
+
 ## Version 6.1.3
 
 * Rework the status commands of the ecctool for more details - Issue #946
--- a/docs/METRICS.md
+++ b/docs/METRICS.md
@@ -26,6 +26,30 @@ Metrics reported using different formats may look differently,
 for reference please refer to ecChronos metrics section below.
 Metrics reported using `file` will be written in CSV format.
 
+## REST API
+
+When HTTP reporting is enabled, metrics are also available via the REST endpoint `/metrics`.
+
+### Retrieving all metrics
+
+To retrieve all available metrics, call the endpoint without parameters:
+```
+GET /metrics
+```
+
+### Filtering specific metrics
+
+To retrieve only specific metrics, use the `name[]` query parameter:
+```
+GET /metrics?name[]=repaired.ratio&name[]=node.repaired.ratio
+```
+
+### Content type
+
+The endpoint supports both Prometheus and OpenMetrics formats:
+- Default: `text/plain; version=0.0.4; charset=utf-8` (Prometheus format)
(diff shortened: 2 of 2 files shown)


jwaeab (changes requested):
Update CHANGELOG.md.

The changed behavior should be described in docs / METRICS as well.

PR #1267, opened 2025-12-01
Agent/1189v2

paulchandler:
Loads of Debugging messages added.
Updates to ensure the schedule threads are created correctly when new nodes are added, and to ensure any nodes added during the startup of Echonos have schedule threads created correctly.


jwaeab (commented):
Are all these debugs kind of temporary for the testing, or are the meant to be kept in the code base?

jwaeab (approved):

Issue #1264, opened 2025-11-28
Jolokia Connection Failure with External Deployment (IPv6 and IPv4)


cezarpaulo16:
## Description:

**1- Does the problem persist?**
Yes, the problem persists consistently in both IPv6 and IPv4 environments when Jolokia JMX is enabled.

**2- Describe how to reproduce the issue:**
1. Deploy ecChronos with 2 external DCs environment (IPv6 or IPv4)
2. Configure ecChronos with Jolokia enabled for JMX connections
3. Trigger repair operations
4. Observe JMX connection failures in ecChronos logs
5. Verify repair jobs remain stuck in `IN_QUEUE` status

## Detailed description:

**• What is happening?**
ecChronos fails to establish JMX connections to Cassandra nodes when Jolokia is enabled, affecting both IPv6 and IPv4 environments.

**• What did you expect to happen?**
JMX connections should succeed with both IPv6 and IPv4 addresses when Jolokia is enabled, allowing repair operations to proceed normally. Repair jobs should transition from `IN_QUEUE` to active execution.

**• What have you tried?**
- Tested both IPv6 and IPv4 deployments

**• What version of ecChronos are you using?**
Latest version from agent/master branch (agent mode deployment)

**• Was the problem detected during an upgrade or downgrade procedure?**
No.

**• Have you checked the ecChronos documents?**
Yes.

**• What do YOU think is the issue?**
Two separate but related issues:
1. **IPv6**: Maybe URL formatting problem
2. **IPv4**: Connection establishment or validation failures (It is using IP directly, we had a problem with TLS when using IP instead of host name, but I don't know if that is the cause as the erro only says "because of null")

**• Anything else that is worth being noted?**

## Log Evidence:

**IPv6 - Malformed URL:**
```json
{
  "message": "[pool-8-thread-1] [com.ericsson.bss.cassandra.ecchronos.connection.impl.builders.DistributedJmxBuilder] Failed to create JMX connection with node 908f9669-c94e-4296-b547-56d761abf98f because of Bad URL path: b70:820d:2812:0:0:0:846:7778/jolokia/",
  "severity": "error",
  "timestamp": "2025-11-27T19:20:40.634Z"
}
```

**IPv4 - Connection/Validation Failure:**
```json
{
[...]

PR #1284, opened 2025-12-08
Fix MalformedURLException for IPv6 addresses

tommystendahl:
Fix part of #1264, an updated Jolokia version is still needed to fully support IPv6 addresses.


VictorCavichioli (approved):
LGTM

PR #1362, opened 2026-02-02
Docs: clarify Jolokia connectivity and node address resolution

0rlych1kk4:
This PR adds a short troubleshooting note clarifying that ecChronos derives node addresses from Cassandra topology via the Java driver, and therefore expects Jolokia to be exposed on the same network address as Cassandra’s
native CQL port.

This helps avoid confusion in Kubernetes or external load balancer setups where reverse DNS or hostname resolution may resolve to unexpected addresses.

### DIFF
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -31,6 +31,15 @@ Change the level to the following (and then restart ecChronos):
     <appender-ref ref="FILE_DEBUG" />
 </root>
 ```
+### Jolokia connectivity and node address resolution
+
+ecChronos derives node addresses directly from Cassandra topology via the Java driver.
+As a result, Jolokia must be exposed on the same network address as Cassandra’s native CQL port.
+
+ecChronos does not perform independent hostname or reverse DNS resolution for Jolokia.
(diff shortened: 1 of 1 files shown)


0rlych1kk4:
> @0rlych1kk4 Thanks for your contribution. This should not be done on the master branch, the current ecChronos version does not use jolokia and is a sidecar solution. Jolokia is used in the development of the new architecture ecChronos on the agent/master branch. So far there is only a bete release of this work.

@tommystendahl Thanks for pointing that out. I missed the distinction between the current master and the agent-based architecture. Understood that Jolokia applies only to the new agent/master work and not the current ecChronos master branch. In that case, this clarification doesn’t belong on master.
I can close this PR, or retarget it to the appropriate branch if you think the note would be useful there. Happy to follow your preference

0rlych1kk4:
> @0rlych1kk4 I think your changes is applicable in the agent/master branch so if could target the PR.

Thanks @tommystendahl, that works for me. I’ll retarget this PR to the agent/master branch and keep it scoped to documentation only.

tommystendahl:
@0rlych1kk4 
I think the best approach would be to do a new PR.

PR #1373, opened 2026-02-07
test: verify configured TLS cipher suites propagate to SSLEngine

0rlych1kk4:
This tightens TestReloadingCertificateHandler to check that configured TLS cipher suites are actually propagated to the created SSLEngine.
The test uses Assume to skip when the configured suites are not supported by the current JDK/provider, to avoid environment-specific failures.

Verified locally:
mvn -pl application test

Related to #1002


tommystendahl:
Hi,
Thanks for your PR. Unfortunately it looks like you base your work on the wrong branch, it looks like you are merging the agent/master branch into the master branch. Could you please base your work on the master branch.

PR #1480, opened 2026-04-10
Improve parallelism of Lock operations

VictorCavichioli:


tommystendahl (dismissed):

tommystendahl (dismissed):

core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/locks/LockCache.java
```diff
@@ -85,7 +90,7 @@ public DistributedLock getLock(final String dataCenter,
 
     private void throwCachedLockException(final LockException e) throws LockException
     {
-        LOG.debug("Encountered cached locking failure, throwing exception", e);
+        LOG.info("[DIAG] Lock blocked by cached failure: {}", e.getMessage());
```
jwaeab, line 93: What is the idea with these [DIAG] log entries? Are they temporary for this PR? I mean, going from DEBUG to INFO and attaching a DIAG tag...

tommystendahl, line 93: Please, don't do e.getMessage(). It will only print the message, it dosen't tell you what the exception was. Remove to {} and do just e the way it was before, that will give you the expetion, the message and the stack trace.

tommystendahl (changes requested):

PR #1511, opened 2026-04-28
Make HangPreventingTask check actual repair status

### LINKED ISSUE #1501: Under certain circumstances on-demand jobs can get stuck
When running on-demand jobs ("run-repair"), the list of jobs can get stuck if, for instance, a "completed" has been dropped over the JMX connection. If so, the current implementation keeps on waiting for it as long as the node is up. Eventually it will abort after 40 minutes (hardcoded 4x10 minutes).

Since on-demand works sequentially, all jobs after the hanging one will remain until it is aborted.

The solution to this should perhaps be (still checking if the node is up) to see if jobs are still active. If so, it can be assumed it is still working and is just slow. If, however, it is reported the repair is not ongoing, it is most likely due to a dropped response, and we can log it and move on to the next job.
The timeouts in between these checks could be much smaller, perhaps 1 minute. There might still be a need for an upper ceiling - in which the job monitoring is still aborted.

Note: This behavior is observed when Cassandra still actually does the repairs successfully. Therefore the assumption is that "completed" notifications have been dropped.

jwaeab:
Closes #1501

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Version 6.1.8
 
+* Make HangPreventingTask check actual repair status - Issue #1501
+
 ## Version 6.1.7
 
 ## Version 6.1.6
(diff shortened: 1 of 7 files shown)


VictorCavichioli (approved):
LGTM

tommystendahl (approved):

PR #1574, opened 2026-05-20
Refactor DistributedJmxBuilder to Reduce Cyclomatic Complexity

### LINKED ISSUE #1573: Refactor DistributedJmxBuilder to Reduce Cyclomatic Complexity
Story Description:

The DistributedJmxBuilder.reconnect() method currently has high cyclomatic complexity because it handles host resolution, protocol selection (Jolokia vs RMI), connection creation, timeout management, TLS/credentials configuration, and connection verification all in a single method. This makes the code difficult to maintain, test, and extend.

The method contains a large if/else block that branches on isJolokiaEnabled, with each branch containing ~40 lines of inline logic including nested try/catch blocks. Additionally, helper methods like createJMXEnv() mix Jolokia-specific SSL property setting with RMI socket factory configuration in the same method, further increasing complexity.

The proposed improvement is to apply the Strategy Pattern by extracting the connection logic into a JmxConnectionStrategy interface with two implementations: JolokiaConnectionStrategy and RMIConnectionStrategy. Each strategy encapsulates its own host resolution, connection creation, environment setup, and protocol-specific behavior (e.g., Jolokia timeout with ExecutorService). The decision of which strategy to use moves to the composition root (AgentJmxConnectionProvider), and reconnect() becomes a simple delegation call.

**Relevant files:**

1. connection.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/connection/impl/builders/DistributedJmxBuilder.java
2. connection/src/main/java/com/ericsson/bss/cassandra/ecchronos/connection/JmxConnectionStrategy.java (new interface)
3. connection.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/connection/impl/builders/utils/JolokiaConnectionStrategy.java (new)
4. connection.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/connection/impl/builders/utils/RMIConnectionStrategy.java (new)
5. connection.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/connection/impl/builders/utils/ConnectionUtils.java (new — shared utilities)
6. application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/providers/AgentJmxConnectionProvider.java

**Acceptance Criteria:**

- [ ] DistributedJmxBuilder.reconnect() delegates to JmxConnectionStrategy.connect(Node) without any protocol-specific branching.
- [ ] JmxConnectionStrategy interface defines ConnectionResult connect(Node node) returning a record with JMXConnector and JMXServiceURL, eliminating mutable shared state for thread-safety.
- [ ] JolokiaConnectionStrategy encapsulates Jolokia-specific logic: protocol selection, timeout via ExecutorService, MBeanServerConnection verification, and PEM certificate system properties.
- [ ] RMIConnectionStrategy encapsulates RMI-specific logic: JMX URL formatting, SslRMIClientSocketFactory configuration, and TLS system properties.
- [ ] ConnectionUtils extracts shared concerns: host resolution (broadcast address, IP translation, reverse DNS, IPv6 formatting), credential/TLS supplier access, and static utilities (isConnected, closeQuietly).
- [ ] The strategy is selected once at construction time in AgentJmxConnectionProvider based on jolokiaConfig.isEnabled(), not re-evaluated per connection.
- [ ] Port configuration is provided at build time to each strategy (from ecc.yml), removing the per-node CQL query for JMX port (getJMXPort).

**Definition of Done:**

- [ ] All existing integration tests pass (standalone-integration).
- [ ] No checkstyle violations introduced.
- [ ] Javadoc present on all public classes and methods.
- [ ] DistributedJmxBuilder no longer has the PMD GodClass suppression.

Notes:

- The getJMXPort CQL query (system_views.system_properties) is removed since all nodes in a cluster use the same JMX port configured in cassandra-env.sh/jvm.options. The port is now a static configuration value in ecc.yml.
- ConnectionResult is defined as a record inside the JmxConnectionStrategy interface to solve the thread-safety issue where the previous jmxUrl field could be overwritten by concurrent reconnect calls during createConnections.

Related with #652

### LINKED ISSUE #1578: Deprecate JMX Port Discover to use port field in ecc.yml
**Story Description:**

Currently, the JMX RMI port for connecting to Cassandra nodes is discovered at runtime via a CQL query to **system_views.system_properties** on each node. This approach introduces unnecessary overhead (N queries for N nodes that all return the same value), is fragile when nodes are down, and assumes access to **system_views** which may not always be available.

In practice, all Cassandra nodes in a cluster use the same JMX port, configured statically in cassandra-env.sh or jvm.options. The port should be a static configuration value in ecc.yml under the jmx connection section, consistent with how the Jolokia port is already configured.

The change adds a port field to the jmx section in ecc.yml with a default value of 7199, and removes the per-node CQL-based port discovery (JmxPortDiscovery).

**Relevant files:**

- application/src/main/resources/ecc.yml
- application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/config/connection/DistributedJmxConnection.java

**Acceptance Criteria:**

- A port field exists under the jmx connection section in ecc.yml with default value 7199.
- The configured port is used for all RMI JMX connections without querying individual nodes.
- The JmxPortDiscovery CQL-based port lookup is removed.

**Definition of Done:**

- Configuration is documented in ecc.yml with comments explaining the expected value and its relationship to Cassandra's cassandra.jmx.remote.port / cassandra.jmx.local.port.
- Existing integration tests pass without modification.

**Notes:**

- This assumes all nodes managed by the same ecChronos instance use the same JMX port. Heterogeneous port configurations across nodes in the same cluster are not a supported scenario.
- Default value 7199 matches Cassandra's default JMX port.
- Related with #652

VictorCavichioli:
Closes #1573 
Closes #1578


connection.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/connection/impl/builders/utils/ConnectionUtils.java
```diff
@@ -0,0 +1,251 @@
...
+        }
+        if (myIpTranslator.isActive())
+        {
+            host = myIpTranslator.getInternalIp(host);
+        }
+        if (myReverseDNSResolution)
+        {
+            host = ReverseDNS.fromHostString(host);
+        }
+        if (host.contains(":") && !host.startsWith("[") && !host.endsWith("]"))
```
jwaeab: Doesn't this miss the cases where you have one square bracket missing? Perhaps the code should be more along the lines of:

if (host.contains(":"))
{
    if (!host.startsWith("["))
        host = "[" + host;
    if (!host.endsWith("]"))
        host = host + "]";
}

jwaeab (changes requested):

connection.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/connection/impl/builders/utils/ConnectionUtils.java
jwaeab: ... or perhaps use something like using Java's InetAddress / Inet6Address?

application/src/main/resources/ecc.yml
```diff
@@ -143,6 +143,18 @@ connection:
       time: 45
       unit: MINUTES
   jmx:
+    ##
+    ## The JMX port used for native RMI connections to Cassandra nodes.
+    ## This value must match the port configured in Cassandra's cassandra-env.sh
+    ## or jvm.options (cassandra.jmx.remote.port / cassandra.jmx.local.port).
+    ## All nodes in the cluster are expected to use the same JMX port.
```
tommystendahl: I think different ecchronos instances can use different ports, so it is only the nodes controlled by the same instance that needs to use the same port. Probably a rare setup but still possible,

CHANGES.md
```diff
@@ -2,6 +2,7 @@
 
 ## Version 1.0.0 (Not yet Released)
 
+* Refactor DistributedJmxBuilder to Reduce Cyclomatic Complexity - Issue #1573
```
tommystendahl, line 6: I think the focus of the change log is changes affecting the user so internal refactoring is probably not that relevant. But that you changed from jmx port discovery to setting the port in the config is probably much more relavent.

connection.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/connection/impl/builders/utils/JmxHostResolver.java
```diff
@@ -0,0 +1,75 @@
...
+        }
+        if (myIpTranslator != null && myIpTranslator.isActive())
+        {
+            host = myIpTranslator.getInternalIp(host);
+        }
+        if (myReverseDNSResolution)
+        {
+            host = ReverseDNS.fromHostString(host);
+        }
+        if (host.contains(":") && !host.startsWith("[") && !host.endsWith("]"))
```
jwaeab: Still sensitive to malformed addresses; e.g. "[::1"?

jwaeab (changes requested):

jwaeab (dismissed):

tommystendahl (dismissed):

jwaeab (approved):

PR #1588, opened 2026-05-22
Avoid Redundant Usage of LongTokenRange by Using Cache

### LINKED ISSUE #1580: Implement Flyweight Caching for LongTokenRange Instances
## Story Description

ecChronos currently recreates all `VnodeRepairState` objects during every repair state update cycle (~30 seconds). Each `VnodeRepairState` contains a `LongTokenRange` instance representing a vnode token range. Since the cluster topology remains stable between cycles, the same token ranges are repeatedly instantiated thousands of times across the application lifecycle.

In the current environment, ecChronos manages approximately 408 tables with ~54 vnodes each, resulting in roughly 22,752 `LongTokenRange` allocations per cycle. However, the cluster itself only contains approximately 324 unique token ranges. This causes unnecessary heap usage, excessive object duplication, and avoidable garbage collection pressure.

The `LongTokenRange` class located in `core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/state/LongTokenRange.java` is currently an immutable value object containing only two `long` fields and does not implement any caching or interning mechanism.

To reduce allocation churn and memory waste, implement a Flyweight/Interning pattern for `LongTokenRange` using a static thread-safe cache. A factory method such as `LongTokenRange.of(start, end)` should become the preferred creation mechanism across the codebase, replacing direct constructor invocations where applicable.

The implementation must ensure thread safety, preserve immutability, and avoid cache key collisions. Prefer using the object itself as the cache key leveraging the existing `equals` and `hashCode` implementations instead of relying on computed primitive hash keys.

---

## Acceptance Criteria

- A static cache mechanism is implemented in `LongTokenRange` to reuse identical token range instances across the JVM lifecycle.
- A factory method (`LongTokenRange.of(start, end)`) is introduced and documented as the preferred object creation mechanism.
- All relevant `new LongTokenRange(start, end)` instantiations across the codebase are replaced with the factory method where applicable.
- The implementation avoids hash collision risks by using a safe cache key strategy.
- The cache implementation is thread-safe and supports concurrent access without synchronization issues.
- Existing immutability guarantees of `LongTokenRange` are preserved.
- Unit tests verify that identical ranges return the same object reference instance (`==` equality).
- Existing functionality and serialization/deserialization behavior remain backward compatible.

---

## Definition of Done

- Implementation is completed and merged into the target branch.
- All unit and integration tests pass successfully.
- New unit tests covering cache reuse and reference equality are added.
- Javadocs or inline documentation explain the Flyweight caching behavior and intended usage of the factory method.
- Static analysis and code quality checks pass without introducing new warnings or violations.

---

## Notes

- The cache size is expected to remain very small (~324 entries) because token ranges are topology-bound.
- Old cached ranges might be updated after topology changes during to the rebalance of vnodes.
- Prefer using `ConcurrentHashMap<LongTokenRange, LongTokenRange>` with `putIfAbsent` or equivalent APIs to avoid collision-related issues.
- The public constructor may remain available for backward compatibility, testing, or deserialization purposes, but the factory method should be the recommended entry point.

VictorCavichioli:
Closes #1580

### DIFF
--- a/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/state/LongTokenRange.java
+++ b/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/state/LongTokenRange.java
@@ -16,6 +16,10 @@ package com.ericsson.bss.cassandra.ecchronos.core.state;
 
 import java.math.BigInteger;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 
 /**
  * A representation of a token range in Cassandra.
@@ -26,7 +30,9 @@ import java.util.concurrent.ConcurrentHashMap;
 @SuppressWarnings("VisibilityModifier")
 public class LongTokenRange
 {
-    private static final ConcurrentHashMap<LongTokenRange, LongTokenRange> CACHE = new ConcurrentHashMap<>();
+    private static final Cache<LongTokenRange, LongTokenRange> CACHE = Caffeine.newBuilder()
+        .expireAfterAccess(60, TimeUnit.MINUTES)
+        .build();
 
     private static final int HASH_THIRTYONE = 31;
     private static final int HASH_THIRTYTWO = 32;
@@ -50,9 +56,8 @@ public class LongTokenRange
      */
     public static LongTokenRange of(final long start, final long end)
     {
-        LongTokenRange candidate = new LongTokenRange(start, end);
-        LongTokenRange cached = CACHE.putIfAbsent(candidate, candidate);
-        return cached != null ? cached : candidate;
+        LongTokenRange key = new LongTokenRange(start, end);
+        return CACHE.get(key, k -> k);
     }
 
     /**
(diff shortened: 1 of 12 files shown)


tommystendahl (dismissed):

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/state/LongTokenRange.java
```diff
@@ -15,13 +15,25 @@
...
  * A representation of a token range in Cassandra.
+ * <p>
+ * Prefer using {@link #of(long, long)} to obtain instances, as it leverages an internal
+ * flyweight cache to avoid redundant allocations for the same token range.
  */
 @SuppressWarnings("VisibilityModifier")
 public class LongTokenRange
 {
+    private static final int DEFAULT_TTL_IN_MINUTES = 60;
+    private static final Cache<LongTokenRange, LongTokenRange> CACHE = Caffeine.newBuilder()
```
tommystendahl, line 33: Why did you choose to switch to Caffeine?

tommystendahl (approved):

core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/state/LongTokenRange.java
VictorCavichioli: AS we discussed, the concern is that it scales linear with the number of nodes that the instance manages, so in terms of space it can grow a lot, if the collection does not have a TTL, we would maintain all the unused token ranges for a long time and we don't know how much data as all the variables related with token range comes from customer clustes, but we know that the number of object would be more or less number_of_nodes * num_tokens * factor_of_subranges

PR #1593, opened 2026-05-25
Reduce Memory Footprint of VnodeRepairState

### LINKED ISSUE #1586: Reduce Memory Footprint of VnodeRepairState
## Story Description

`VnodeRepairState` is currently the most instantiated class in the ecChronos heap, with approximately 21,888 live instances consuming over 1MB of memory. Each instance stores vnode repair metadata including token range, replica set, repair timestamps, and a precomputed repair duration.

The class currently stores the `repairTime` value as a dedicated field even though the value is directly derivable from `finishedAt - startedAt`. This introduces unnecessary per-instance memory overhead across thousands of objects.

Because `VnodeRepairState` exists in very large quantities and participates heavily in repair state recomputation cycles, even small reductions in object size can produce meaningful overall heap savings and reduce cache pressure.

The implementation should remove the redundant `repairTime` field and compute the value dynamically when requested. Additionally, the class structure should be evaluated for possible migration to a Java 17 `record` to simplify boilerplate and potentially improve JVM optimization opportunities.

The implementation must preserve immutability semantics and avoid changing observable repair behavior.

Affected file:

- `core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/state/VnodeRepairState.java`

---

## Acceptance Criteria

- The redundant `repairTime` field is removed from `VnodeRepairState`.
- `getRepairTime()` computes the repair duration dynamically from existing timestamp fields.
- Existing repair time semantics remain unchanged.
- Existing immutability guarantees are preserved.
- Equality and hash code behavior remain correct after the field removal.
- Existing APIs remain backward compatible if the class remains a standard class implementation.
- If migrated to a Java record, all affected usages across the codebase are updated accordingly.
- Unit tests validate repair time calculations for repaired and unrepaired vnode states.
- Unit tests validate equality and hash code behavior after the refactor.
- Memory footprint reduction is verified through profiling or object layout analysis.

---

## Definition of Done

- Implementation is completed and merged into the target branch.
- All unit and integration tests pass successfully.
- Static analysis and code quality checks pass without introducing new warnings or violations.
- JOL or equivalent object layout analysis confirms reduced object footprint.
- Documentation or inline comments explain the rationale for computing repair time dynamically.
- Any required refactoring caused by record migration is completed successfully.

---

## Notes

- Computing `finishedAt - startedAt` on demand is significantly cheaper than permanently storing a redundant field across tens of thousands of objects.
- `getRepairTime()` is not a hot-path operation and the additional subtraction cost is negligible.
- If the class is converted to a Java record, method naming conventions will change (`getTokenRange()` → `tokenRange()`), which may require broader refactoring.
- Verify that `VnodeRepairState` is not used in serialization-sensitive contexts before removing fields or converting to a record.
- Records are a good fit for immutable value objects and automatically provide `equals`, `hashCode`, and `toString` implementations.

VictorCavichioli:
Closes #1586

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Version 1.0.0 (Not yet Released)
 
+* Reduce Memory Footprint of VnodeRepairState - Issue #1586
 * Pre-size LinkedHashMap Allocation in VnodeRepairStatesImpl.Builder - Issue #1587
 * Optimize VnodeRepairStatesImpl combineWithRepairedAt Allocations - Issue #1582
 * Implement ReplicaSet Interning for VnodeRepairState Instances - Issue #1581
--- a/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/state/VnodeRepairState.java
+++ b/core/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/state/VnodeRepairState.java
@@ -30,7 +30,6 @@ public class VnodeRepairState
     private final ImmutableSet<DriverNode> myReplicas;
     private final long myStartedAt;
     private final long myFinishedAt;
-    private final long myRepairTime;
 
     /**
      * Constructor.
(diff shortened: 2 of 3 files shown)


tommystendahl (dismissed):

tommystendahl (approved):

tommystendahl:
@VictorCavichioli Was there a reason you didn't change to a record?

VictorCavichioli:
> @VictorCavichioli Was there a reason you didn't change to a record?

According to this https://docs.oracle.com/en/java/javase/17/language/records.html? there's no exactly memory improvement from java side when we use records, and looks like VnodeRepairState is more a utility class then a record

PR #1617, opened 2026-05-29
Fix RepairStateSnapshot retention by in-flight repair tasks

### LINKED ISSUE #1616: RepairStateSnapshot retention by in-flight repair tasks causes continued memory growth
## Problem

Despite the lazy repair state fix (#1599), `RepairStateSnapshot` objects continue to accumulate in the heap. After 40 hours of operation:

- **14,950** `RepairStateSnapshot` instances (expected: 408, one per table)
- **690,804** `VnodeRepairState` instances (~60 MB)
- **681,524** `LongTokenRange` instances (~22 MB)
- Growing at ~270 snapshots/hour and ~12K VnodeRepairState/hour

The lazy state fix reduced the *creation* rate (from ~89 snapshots/table in beta7 to ~37/table), but old snapshots are still retained longer than necessary.

## Root Cause

`TableRepairJob.iterator()` creates `RepairGroup` tasks that hold references to `ReplicaRepairGroup` objects from the current snapshot:

```java
for (ReplicaRepairGroup replicaRepairGroup : repairStateSnapshot.getRepairGroups()) {
    RepairGroup.Builder builder = RepairGroup.newBuilder()
            .withReplicaRepairGroup(replicaRepairGroup)  // pins snapshot
            ...
}
```

`ReplicaRepairGroup` is stored in the snapshot's `ImmutableList<ReplicaRepairGroup>`, so holding a reference to any `ReplicaRepairGroup` pins the **entire** `RepairStateSnapshot` and its full `VnodeRepairStates` collection (~46 `VnodeRepairState` objects per snapshot).

With repairs taking minutes to complete and the scheduler creating new snapshots every 30 seconds, multiple generations of snapshots remain pinned by in-flight tasks until those tasks finish.

### Retention chain:
```
RepairGroup (86 active)
  → ReplicaRepairGroup (member of snapshot's ImmutableList)
    → RepairStateSnapshot (pinned, cannot be GC'd)
      → VnodeRepairStatesImpl
        → ImmutableList<VnodeRepairState> (~46 objects)
          → LongTokenRange + ImmutableSet<DriverNode> per state
```

## Proposed Fix

Decouple repair tasks from the snapshot object graph. `RepairGroup` only needs the token ranges and replica set from its `ReplicaRepairGroup` — it does not need a reference back to the snapshot.

**Option A:** Copy the needed data out of `ReplicaRepairGroup` when creating the `RepairGroup`, breaking the reference to the snapshot's list.

**Option B:** Make `ReplicaRepairGroup` a standalone object that is not stored inside the snapshot's `ImmutableList` (i.e., create independent copies for the task).

## Evidence

| Metric | 17h | 40h | Growth rate |
|--------|-----|-----|-------------|
| RepairStateSnapshot | 8,696 | 14,950 | ~270/hour |
| VnodeRepairState | 412,368 | 690,804 | ~12K/hour |
| Active RepairGroup | — | 86 | — |
| Active VnodeRepairTask | — | 180 | — |

## Impact

~82 MB of heap retained by old snapshots (VnodeRepairState + LongTokenRange). With the proposed fix, only the current 408 snapshots would be retained (~2 MB), freeing ~80 MB.

tommystendahl:
Create a defensive copy of ReplicaRepairGroup when constructing RepairGroup, breaking the reference chain from in-flight tasks back to the snapshot's ImmutableList. This allows old RepairStateSnapshot objects to be GC'd immediately when replaced, instead of being pinned until all repair tasks referencing their members complete.

Frees ~80MB of retained heap (VnodeRepairState + LongTokenRange).

Closes #1616

### DIFF
--- a/core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/RepairGroup.java
+++ b/core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/RepairGroup.java
@@ -89,8 +89,8 @@ public class RepairGroup extends ScheduledTask
 
         myRepairConfiguration = Preconditions
                 .checkNotNull(builder.myRepairConfiguration, "Repair configuration must be set");
-        myReplicaRepairGroup = Preconditions
-                .checkNotNull(builder.myReplicaRepairGroup, "Replica repair group must be set");
+        myReplicaRepairGroup = copyReplicaRepairGroup(Preconditions
+                .checkNotNull(builder.myReplicaRepairGroup, "Replica repair group must be set"));
         myJmxProxyFactory = Preconditions
                 .checkNotNull(builder.myJmxProxyFactory, "Jmx proxy factory must be set");
         myTableRepairMetrics = Preconditions
@@ -265,6 +265,14 @@ public class RepairGroup extends ScheduledTask
             }
         }
         return allowedParticipants;
+    }
+
+    private static ReplicaRepairGroup copyReplicaRepairGroup(final ReplicaRepairGroup original)
+    {
+        return new ReplicaRepairGroup(
+                original.replicas() != null ? new HashSet<>(original.replicas()) : null,
+                original.vnodes() != null ? new ArrayList<>(original.vnodes()) : null,
+                original.lastCompletedAt());
     }
 
     /**
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## Version 1.0.1 (Not yet Released)
 
+* Fix RepairStateSnapshot retention by decoupling RepairGroup from snapshot object graph - Issue #1616
 * Set HttpClient keepalive timeout to 5s to prevent connection pool memory growth - Issue #1614
 * Add MaxMetaspaceSize and ReservedCodeCacheSize to jvm.options - Issue #1610
 * Fix native memory leak: replace per-task executor with shared pool in RepairHangMonitor - Issue #1611


jwaeab (approved):

core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/RepairGroup.java
```diff
@@ -267,6 +267,14 @@ private Set<DriverNode> filterParticipants(final Set<DriverNode> participants, f
...
     }
 
+    private static ReplicaRepairGroup copyReplicaRepairGroup(final ReplicaRepairGroup original)
+    {
+        return new ReplicaRepairGroup(
+                original.replicas() != null ? new HashSet<>(original.replicas()) : null,
+                original.vnodes() != null ? new ArrayList<>(original.vnodes()) : null,
+                original.lastCompletedAt());
+    }
+
```
VictorCavichioli, line 277: I don't know if it would be unnecessary work, but in this case did you consider the extra allocation of this class? Would it increase heap usage? If it makes sense, maybe an implementation similar to what we did with LongTokenRange could help avoid unnecessary allocations with new.

tommystendahl: The idea is that they will be recycled quicker, but I think we need to see the test result before we are sure of that. I think we should merge this so we can get the tests going again. If you like to improve on this after that you are more then welcome.

PR #1619, opened 2026-05-31
Allow One Time Lock Acquisition per Job

### LINKED ISSUE #1618: Implement Batched Lock Acquisition Strategy to Improve Repair Throughput in Large Clusters
**Story Description:**

Currently, the `ScheduleManagerImpl.JobRunTask` acquires and releases a distributed CAS lock **per individual RepairGroup task**. In large clusters (e.g., 60 nodes, 1000+ tables), this creates excessive lock churn: with `DATACENTER` lock type and `locks_per_resource=3`, nodes constantly compete for 3 slots, each holding the lock for only one short repair before releasing and re-competing.

With 60 nodes fighting over 3 slots, the CAS overhead (priority announcement + LWT insert + release) per acquisition dominates scheduling time. The net effect is that the system cannot complete enough repairs within the configured interval, resulting in LATE/OVERDUE statuses.

The proposed improvement changes the lock acquisition strategy from **per-task** to **per-session**: acquire the lock once, execute multiple repair tasks within a configurable time window (default: 5 minutes), then release. This reduces CAS operations by 10-50x without increasing the number of concurrent repairs (still bounded by `locks_per_resource`).

The implementation is located in:
- `core.impl/src/main/java/.../repair/scheduler/ScheduleManagerImpl.java` — batched session logic in `JobRunTask`
- `application/src/main/java/.../config/scheduler/SchedulerConfig.java` — new config properties
- `application/src/main/java/.../spring/ECChronosInternals.java` — wiring new config to builder
- `application/src/main/resources/ecc.yml` — new `session_window` and `cooldown` settings

Additionally, the scheduling tick is changed from fixed-delay (`scheduleWithFixedDelay`) to adaptive self-scheduling: if work was completed, the next tick fires in 1 second instead of waiting the full 30-second interval. Contention backoff is also reduced from [30s-60s] to [15s-30s].

**Acceptance Criteria:**

1. When a node acquires a distributed lock, it holds it for the configured `session_window` duration (default: 5 minutes) and executes multiple repair tasks from multiple jobs in priority order before releasing.

2. The number of concurrent repairs per datacenter remains bounded by `locks_per_resource` — no increase in parallel repair load on Cassandra.

3. Jobs within a batched session are executed in priority order as determined by `DefaultJobComparator` (highest `getRealPriority()` first).

4. After a batched session completes, the node optionally enters a configurable `cooldown` period before attempting to acquire a new lock, allowing fair rotation among nodes.

5. The scheduler tick is adaptive: if the previous tick executed work, the next tick fires in 1 second; if no work was available, it waits the full configured frequency (default: 30 seconds).

6. Contention backoff on lock failure is reduced to [frequency/2, frequency] (default: [15s, 30s]) instead of [frequency, frequency×2].

7. All existing unit tests pass (with test updates reflecting the new batched behavior where lock is acquired once per session instead of per task).

8. Configuration is backward compatible: default values produce behavior similar to the previous implementation (5-minute window means most single-task jobs complete within one session).

**Definition of Done:**

- All unit tests in `core.impl` and `application` modules pass.
- Integration tests in `standalone-integration` pass.
- New configuration properties are documented in `ecc.yml` with comments.
- No increase in Cassandra repair pressure (same `locks_per_resource` concurrency limit).
- Code passes existing checkstyle and PMD rules.

**Notes:**

- The `session_window` default of 5 minutes is well within the lock TTL of 10 minutes, so the existing lock renewal mechanism (every 60s) keeps the lock alive during the session.
- If ecChronos crashes while holding a batched lock, the TTL auto-expires it (10 min). No data corruption — repairs are idempotent.

Related with #652

VictorCavichioli:
Today each RepairTask tries to acquire a lock, in large clusters with 1000+ tables ecchronos would suffer to repair all the things in a feasible time because acquiring a lock involves a CAS query in lock table, because of that I propose to allow a node to run the most priority jobs within a time window before releasing the lock to another node, it would reduce the number of CAS operation and speed up repair jobs without increasing repair concurrency.

**Important**

We still need to discuss appropriate default values for cooldown and session_window. Although we are not increasing repair parallelism, the previous implementation, which used a lock per task, introduced a short gap between repair executions (the time required to release the current lock and acquire the next one). This effectively gave the Cassandra node a brief recovery period between repairs. With the new approach, however, subsequent repair tasks may run back-to-back within the same session_window, potentially increasing the sustained load on the node.

I'm planning to benchmark this change in a week to verify current default values, but I'm not too concerned tho because repair tasks are small and default session_window is 5 minutes , but we need to test it anyway

Closes #1618

### DIFF
--- a/core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/scheduler/ScheduleManagerImpl.java
+++ b/core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/scheduler/ScheduleManagerImpl.java
@@ -17,14 +17,13 @@ package com.ericsson.bss.cassandra.ecchronos.core.impl.repair.scheduler;
 import com.datastax.oss.driver.api.core.metadata.Node;
 import com.ericsson.bss.cassandra.ecchronos.connection.DistributedNativeConnectionProvider;
 import com.ericsson.bss.cassandra.ecchronos.core.impl.locks.CASLockFactory;
-import com.ericsson.bss.cassandra.ecchronos.core.locks.LockFactory;
+import com.ericsson.bss.cassandra.ecchronos.utils.exceptions.LockException;
 import com.ericsson.bss.cassandra.ecchronos.core.repair.scheduler.RunPolicy;
 import com.ericsson.bss.cassandra.ecchronos.core.repair.scheduler.ScheduleManager;
 import com.ericsson.bss.cassandra.ecchronos.core.repair.scheduler.ScheduledJob;
 import com.ericsson.bss.cassandra.ecchronos.core.repair.scheduler.ScheduledTask;
 import java.io.Closeable;
 import java.util.Collection;
-import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
@@ -234,7 +233,7 @@ public final class ScheduleManagerImpl implements ScheduleManager, Closeable
     @VisibleForTesting
     public void run(final UUID nodeID)
     {
-        myRunTasks.get(nodeID).run();
+        myRunTasks.get(nodeID).runOnce();
     }
 
     /**
@@ -285,6 +284,12 @@ public final class ScheduleManagerImpl implements ScheduleManager, Closeable
         @Override
         public void run()
         {
+            boolean hadWork = runOnce();
+            reschedule(hadWork);
+        }
+
+        private boolean runOnce()
+        {
             boolean hadWork = false;
             try
             {
@@ -292,7 +297,7 @@ public final class ScheduleManagerImpl implements ScheduleManager, Closeable
                 if (System.currentTimeMillis() < myCooldownUntil)
                 {
                     LOG.debug("Node {} in cooldown until {}", nodeID, myCooldownUntil);
-                    return;
+                    return false;
                 }
                 if (myQueue.get(nodeID) != null)
                 {
@@ -307,10 +312,7 @@ public final class ScheduleManagerImpl implements ScheduleManager, Closeable
             {
                 LOG.error("Exception while running job in node {}", nodeID, e);
             }
-            finally
-            {
-                reschedule(hadWork);
-            }
+            return hadWork;
         }
 
         private void reschedule(final boolean hadWork)
@@ -362,11 +364,8 @@ public final class ScheduleManagerImpl implements ScheduleManager, Closeable
                 if (validate(next))
                 {
                     currentExecutingJobs.put(nodeID, next);
-                    if (tryRunBatchedSession(next))
-                    {
-                        hadWork = true;
-                        break;
-                    }
+                    hadWork = runSession(next);
+                    break;
                 }
             }
             currentExecutingJobs.remove(nodeID);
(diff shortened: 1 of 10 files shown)


core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/scheduler/ScheduleManagerImpl.java
```diff
@@ -342,73 +387,191 @@ private boolean validate(final ScheduledJob job)
...
+            Iterator<ScheduledTask> taskIterator = firstJob.iterator();
+            if (!taskIterator.hasNext())
+            {
+                return false;
+            }
 
-            int index = 0;
-            for (ScheduledTask task : next)
+            ScheduledTask firstTask = taskIterator.next();
+            try (LockFactory.DistributedLock sessionLock = firstTask.getLock(myLockFactory, nodeID))
```
tommystendahl: Don't we take the for the replica set involved in the ScheduledTask? If we reuse the lock for the next Scheduled task it could be for a different replica set.

tommystendahl:
Only one test was updated. There's no test verifying:
  
  - Multiple jobs execute within a single session
  - Lock resource compatibility across jobs
  - Cooldown behavior
  - Session window timeout mid-execution

core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/scheduler/SessionLockPool.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        Map<RepairResource, LockFactory.DistributedLock> newlyAcquired = new HashMap<>();
+        for (RepairResource resource : missing)
+        {
+            try
+            {
+                LockFactory.DistributedLock lock = myLockFactory.tryLock(
+                        resource.getDataCenter(),
+                        resource.getResourceName(1),
```
tommystendahl: Should this be hard coded to 1?

core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/scheduler/SessionLockPool.java
```diff
@@ -0,0 +1,139 @@
...
+                        task.getPriority(),
+                        new HashMap<>(),
```
tommystendahl: And should this be a empty map?

0rlych1kk4:
Hi @VictorCavichioli and @tommystendahl ,

I noticed the review feedback around missing test coverage for the one-time lock acquisition behavior.

I can help by adding focused tests for:
- multiple jobs executing within a single session window
- lock resource compatibility across jobs
- cooldown behavior
- session window timeout while execution is ongoing

I will keep the first contribution test-only unless the tests expose a behavior gap.

core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/scheduler/SessionLockPool.java
VictorCavichioli: No, this should be using RepairLcokFactory instead, thanks to notice this, going to update before leaving

core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/scheduler/SessionLockPool.java
VictorCavichioli: same as above

VictorCavichioli: same as above

VictorCavichioli:
> Hi @VictorCavichioli and @tommystendahl ,
> 
> I noticed the review feedback around missing test coverage for the one-time lock acquisition behavior.
> 
> I can help by adding focused tests for:
> 
> * multiple jobs executing within a single session window
> * lock resource compatibility across jobs
> * cooldown behavior
> * session window timeout while execution is ongoing
> 
> I will keep the first contribution test-only unless the tests expose a behavior gap.

Hi!

Your contributions are always welcome :D

tommystendahl (approved):

0rlych1kk4:
Thanks @VictorCavichioli ! I appreciate it. I’m happy to continue contributing where I can.
I’ll create a separate test-focused PR from the latest master branch to cover the session lock behavior added in this change.

0rlych1kk4:
I reviewed the merged change and will start with a small test-focused PR around the new scheduler session_window and cooldown behavior.
I’ll first cover the configuration/default values and one core session behavior, then can follow up with additional coverage for compatibility and cooldown edge cases if useful.

0rlych1kk4:
Hi @VictorCavichioli and @tommystendahl ,

I’ve opened the follow-up test-focused PR here: #1660 

The focused session lock pool tests pass locally. The failing broader CI checks appear to be unrelated Cassandra/Testcontainers readiness issues, as `core.impl` builds successfully and none of the failures reference `TestSessionLockPool`.

PR #1646, opened 2026-06-08
Fixing misc test cases

jwaeab:


core/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/TestCASLockFactory.java
```diff
@@ -583,4 +596,10 @@ private long getWriteCountFromTableStats(String tableName) throws UnsupportedOpe
         return writeCount;
     }
 
+    private void awaitSchemaAgreement()
+    {
+        mySession.checkSchemaAgreement();
```
tommystendahl: Do we need to check is checkSchemaAgreement returns true or false?

tommystendahl (dismissed):

core/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/TestCASLockFactory.java
jwaeab: You're right, a simple retry/timeout should probably be added. Do you agree that would suffice?

tommystendahl: Yes, that should be perfect.

tommystendahl (approved):

core/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/TestCASLockFactory.java
jwaeab: Done

PR #1708, opened 2026-07-03
test: cover successful scheduled job not rerun before interval

0rlych1kk4:
Adds regression coverage for ScheduleManagerImpl to verify that a successfully executed scheduled job is not executed again on an immediate subsequent scheduler run.

The job remains queued, but after successful execution it should be PARKED until its configured run interval has elapsed. The test also verifies that the repair lock is acquired only once.

Validation:
mvn -pl core.impl -am -Dtest=TestScheduleManager -Dsurefire.failIfNoSpecifiedTests=false test

### DIFF
--- a/core.impl/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/scheduler/TestScheduleManager.java
+++ b/core.impl/src/test/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/scheduler/TestScheduleManager.java
@@ -119,6 +119,23 @@ public class TestScheduleManager
     }
 
     @Test
+    public void testSuccessfulJobIsNotRunAgainBeforeRunInterval() throws LockException
+    {
+        TestJob job = new TestJob(ScheduledJob.Priority.LOW, 1, nodeID1, "dc1", "resource1");
+        myScheduler.schedule(nodeID1, job);
+
+        when(myLockFactory.tryLock(any(), anyString(), anyInt(), anyMap(), any()))
+                .thenReturn(new DummyLock());
+
+        myScheduler.run(nodeID1);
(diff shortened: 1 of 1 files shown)


0rlych1kk4:
Hi @tommystendahl  and @VictorCavichioli , I checked the failing CI jobs.

The failures appear to be in standalone/binary integration suites, with Docker/container startup or cleanup errors such as a Cassandra container not running and ITSchedules cleanup hitting a null HostStatesImpl. The changed test is only in TestScheduleManager under core.impl, and the core.impl module appears to complete successfully before the integration failures.

Validated locally with:
mvn -pl core.impl -am -Dtest=TestScheduleManager -Dsurefire.failIfNoSpecifiedTests=false test

Please let me know if you would like me to rerun or investigate any specific failing check.

0rlych1kk4:
Hi @tommystendahl  and @VictorCavichioli , just checking if this test-only PR is still useful for the scheduler coverage. The added case is intentionally narrow: it verifies that after a scheduled job runs successfully and remains in the queue, an immediate second scheduler run does not execute the same job again before its configured run interval. It also verifies the repair lock is acquired only once.

I validated the targeted test locally with:

mvn -pl core.impl -am -Dtest=TestScheduleManager -Dsurefire.failIfNoSpecifiedTests=false test

Happy to adjust the assertion or test placement if you prefer a different way to cover this behavior.

tommystendahl (approved):

tommystendahl:
@0rlych1kk4 Thanks for the contibution.

0rlych1kk4:
> @0rlych1kk4 Thanks for the contibution.

Thanks @tommystendahl , I appreciate the review and merge. Glad to contribute to ecChronos.

PR #1713, opened 2026-07-08
Update ecChronos to Spring 4.1.0

### LINKED ISSUE #1711: Upgrade ecChronos to Spring Boot 4
Story Description:
ecChronos is currently built on an older version of the Spring Boot framework, to keep the project aligned with the latest ecosystem the application should be upgraded to Spring Boot 4
Acceptance Criteria:

1. The project builds successfully using Spring Boot 4 and all dependencies are compatible with the new version.
2. All existing unit, integration, and end-to-end tests pass without regressions.
3. Application startup completes successfully without deprecation warnings or runtime errors caused by the migration.
4. All existing REST endpoints, scheduling functionality, persistence layer, metrics, and configuration behave as they did prior to the upgrade.
5. Any deprecated APIs replaced during the migration follow current Spring Boot 4 best practices.
6. Project documentation is updated if configuration changes or new migration requirements are introduced.

Definition of Done:

1. Spring Boot has been upgraded to version 4 across the entire project.
2. All project dependencies are updated to compatible versions.
3. The application compiles and runs successfully.
4. The complete test suite passes in the CI pipeline.
5. Code follows the project's existing coding standards and passes all static analysis checks.
6. Documentation has been updated to reflect any relevant migration changes.
7. The changes have been reviewed and approved through the standard code review process.
8. 

Notes:

1. The upgrade should prioritize maintaining backward-compatible behavior wherever possible.
2. Review the official Spring Boot 4 migration guide for required code and configuration changes.
3. Pay special attention to dependency compatibility, Jakarta namespace changes (if applicable), auto-configuration behavior, and third-party libraries used by ecChronos.
4. If incompatible dependencies are identified, document the issue and propose an appropriate replacement or migration strategy.

VictorCavichioli:
Closes #1711

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Changes
 
+## Version 2.0.0
+
+* Upgrade ecChronos to Spring Boot 4 - Issue #1711
+
 ## Version 1.0.5
 
 * Bump Spring Boot to 3.5.16 and override Tomcat embed to 10.1.56 to fix CVE-2026-53434 - Issue #1706
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # Changes
 
-## Version 2.0.0
+## Version 1.1.0
 
 * Upgrade ecChronos to Spring Boot 4 - Issue #1711
 
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -150,12 +150,12 @@
 
         <!-- Jackson -->
         <dependency>
-            <groupId>com.fasterxml.jackson.dataformat</groupId>
+            <groupId>tools.jackson.dataformat</groupId>
             <artifactId>jackson-dataformat-yaml</artifactId>
         </dependency>
 
         <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
+            <groupId>tools.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
         </dependency>
 
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -93,6 +93,20 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-web</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.tomcat.embed</groupId>
+                    <artifactId>tomcat-embed-el</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.apache.tomcat.embed</groupId>
+                    <artifactId>tomcat-embed-websocket</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.springframework.boot</groupId>
+                    <artifactId>spring-boot-starter-logging</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -93,20 +93,6 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-web</artifactId>
-            <exclusions>
-                <exclusion>
-                    <groupId>org.apache.tomcat.embed</groupId>
-                    <artifactId>tomcat-embed-el</artifactId>
-                </exclusion>
-                <exclusion>
-                    <groupId>org.apache.tomcat.embed</groupId>
-                    <artifactId>tomcat-embed-websocket</artifactId>
-                </exclusion>
-                <exclusion>
-                    <groupId>org.springframework.boot</groupId>
-                    <artifactId>spring-boot-starter-logging</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
 
         <dependency>
--- a/application/pom.xml
+++ b/application/pom.xml
@@ -113,6 +113,12 @@
             <groupId>org.springdoc</groupId>
             <artifactId>springdoc-openapi-starter-webmvc-api</artifactId>
             <scope>provided</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.tomcat.embed</groupId>
+                    <artifactId>tomcat-embed-el</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <!-- Metrics -->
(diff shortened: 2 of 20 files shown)


tommystendahl (dismissed):

CHANGES.md
```diff
@@ -1,5 +1,9 @@
 # Changes
 
+## Version 2.0.0
```
tommystendahl: Do we really need to step to v2? How is this breaking backwards compatibility?

VictorCavichioli: I don't think this breaks backward compatibility, I'm fine on doing a 1.1.0 or something like this

tommystendahl: Why not 1.0.x? All changes are internal, no configuration or api changes or addition as far as I can see.

application/pom.xml
tommystendahl, line 109: Do these exclusions need to be updated?

CHANGES.md
VictorCavichioli: I think 1.x would be a better fit because, as I understand it, 1.0.x should be reserved for bug fixes. In this case, we're introducing two major changes: the Spring migration and the Jackson migration. Additionally, Spring 4.x significantly changes how modules are managed, so this feels more like a minor version update than a patch release.

application/pom.xml
VictorCavichioli: you're probably right, let me fix that

CHANGES.md
tommystendahl: Alright, lets go for 1.1.0 in that case.

application/pom.xml
```diff
@@ -93,20 +93,6 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-web</artifactId>
-            <exclusions>
```
tommystendahl, line 96: Please, don't just remove these exclusions, they are important. I added them to remove unused things from the binary package to reduce the size of the distribution.
My previous comment  was probably unclear, what I meant was that it looks like a lot of package names have changed so we probably have check if these package names need to be updated so we still exclude these things.
We also need to investigate if there are new things in 4.1 that should be excluded.

pom.xml
```diff
@@ -192,7 +191,7 @@
 
     <dependencyManagement>
         <dependencies>
-            <!-- Override tomcat-embed-* to 10.1.56 to fix CVE-2026-53434. Remove once Spring Boot ships with a version that includes the fix. -->
+            <!-- Override tomcat-embed-* to 11.0.24 to fix CVE-2026-53434. Remove once Spring Boot ships with a version that includes the fix. -->
```
tommystendahl, line 194: I found that we have direct usage of the tomcat dependency in our code so we should keep these declarations permanently. To correctly declare direct usage.

tommystendahl (changes requested):

pom.xml
```diff
@@ -446,27 +459,27 @@
...
+                <groupId>tools.jackson.core</groupId>
                 <artifactId>jackson-core</artifactId>
-                <version>${com.fasterxml.jackson.version}</version>
+                <version>${tools.jackson.core.version}</version>
             </dependency>
 
             <dependency>
-                <groupId>com.fasterxml.jackson.core</groupId>
+                <groupId>tools.jackson.core</groupId>
                 <artifactId>jackson-annotations</artifactId>
```
tommystendahl, line 467: I don't think jackson-annotations exists under tools.jackson.core. 

And when I look deeper it seams like tools.jackson.core is still using jackson-annotations from com.fasterxml.jackson.core. In the dependency tree it looks like this:
[INFO] |  \- tools.jackson.core:jackson-databind:jar:3.1.5:compile (version managed from 3.1.5)
[INFO] |     +- com.fasterxml.jackson.core:jackson-annotations:jar:2.21:compile

I think you should change this back to com.fasterxml.jackson.core.

pom.xml
```diff
@@ -249,12 +248,26 @@
...
+            </dependency>
+
             <dependency>
                 <groupId>org.springframework</groupId>
                 <artifactId>spring-context</artifactId>
                 <version>${org.springframework.spring.version}</version>
             </dependency>
 
+
+
```
tommystendahl: Remove extra line

application/pom.xml
tommystendahl: Looks like they have not changed any package names here so if we just put back the exlutions and also add execution for tomcat-embed-el from springdoc-openapi-starter-webmvc-api.

pom.xml
```diff
@@ -446,27 +457,27 @@
...
 
             <dependency>
-                <groupId>com.fasterxml.jackson.dataformat</groupId>
+                <groupId>tools.jackson.dataformat</groupId>
                 <artifactId>jackson-dataformat-yaml</artifactId>
-                <version>${com.fasterxml.jackson.version}</version>
+                <version>${tools.jackson.core.version}</version>
             </dependency>
 
             <dependency>
```
tommystendahl: We should remove the declaration of snakeyaml. We don't ude it directly and Jackson v3 have moved to snakeyaml-engine 3.0.1. Spring boot still use snakeyaml:jar 2.6, but there is no need for us declare the dependency.

pom.xml
tommystendahl, line 267: We should exclude jspecify from spring-context.

      <exclusions>
          <exclusion>
              <groupId>org.jspecify</groupId>
              <artifactId>jspecify</artifactId>
          </exclusion>
      </exclusions>

tommystendahl, line 353: Exclude old Jackson version from the java-driver.

    <exclusion>
      <groupId>com.fasterxml.jackson.core</groupId>
      <artifactId>*</artifactId>
    </exclusion>

pom.xml
```diff
@@ -446,33 +456,27 @@
...
-                <groupId>com.fasterxml.jackson.core</groupId>
+                <groupId>tools.jackson.core</groupId>
                 <artifactId>jackson-core</artifactId>
-                <version>${com.fasterxml.jackson.version}</version>
+                <version>${tools.jackson.core.version}</version>
             </dependency>
 
             <dependency>
-                <groupId>com.fasterxml.jackson.core</groupId>
+                <groupId>tools.jackson.core</groupId>
```
tommystendahl: Thiis must be changed back to com.fasterxml.jackson.core.
tools.jackson.core.jackson-annotations does not exists. tools.jackson.core.jackson-databind V3 is still using com.fasterxml.jackson.core:jackson-annotations V2.

tommystendahl (changes requested):

tommystendahl (approved):

PR #1762, opened 2026-08-27
Fix repair jobs stuck ON_TIME due to empty Jolokia response

### LINKED ISSUE #1740: Repair jobs remain ON_TIME indefinitely when Jolokia endpoint returns empty body due to transient network issues
_[Attach log files and command results. Otherwise, it will not be possible to investigate the problem accurately.]_

Description:

The problem persists as long as the network instability is present. It can be reproduced by introducing network-level packet loss or latency between ecChronos and the Cassandra node's Jolokia endpoint while keeping the Cassandra process itself alive and healthy.

Detailed description:

### What is happening?
Repair jobs remain stuck in ON_TIME status indefinitely when there is a transient network issue between ecChronos and the Cassandra node's Jolokia endpoint. The Jolokia endpoint is reachable (no IOException is thrown) but returns an empty or blank HTTP response body. This causes objectMapper.readValue() in NotificationRunTask.run() to throw a JacksonException, which is not caught by the retry logic inside checkForNotificationsWithRetry (which only catches IOException). As a result, consecutiveFailures increments on every poll cycle (~500ms), reaching the threshold of 5 in approximately 2.5 seconds. signalConnectionFailure() is then called, but if no RepairTask is currently blocked on await()  because the repair never started due to the network issue  the failure notification is dispatched to no consumer and the job status is never updated. At the ScheduleManagerImpl level, runTask() catches all exceptions from task.execute() and silently returns false, causing postExecute(false, task) to be called without ever transitioning the job to a FAILED state. The job is re-queued as ON_TIME and the cycle repeats indefinitely.

### What did you expect to happen?
After a configurable number of consecutive failures to execute a repair task, the job should be transitioned to a FAILED state and the operator should be clearly notified, rather than the job silently looping with ON_TIME status.

### What have you tried?
Observed the behavior through logs, which showed repeated WARN messages from JolokiaNotificationController about transient notification check failures, followed by ERROR logs about connection failure signaling — while the repair job status remained ON_TIME throughout.

### What version of ecChronos are you using?
master version

### Was the problem detected during an upgrade or downgrade procedure?
No

### What do YOU think is the issue (optional)?
Three compounding bugs:

- checkForNotifications does not validate the HTTP response body before returning it, allowing an empty string to propagate to the Jackson deserializer.
- checkForNotificationsWithRetry only retries on IOException, so a JacksonException caused by an empty body bypasses all retry logic entirely.
- ScheduleManagerImpl.runTask() treats all task execution failures as a silent false return with no consecutive failure tracking, so a job that repeatedly fails to start never transitions out of ON_TIME.

### Anything else that is worth being noted (optional)?
The consecutiveFailures counter in NotificationRunTask is also never reset between separate repair executions, meaning failures from a previous network blip can carry over and accelerate the failure threshold for a subsequent healthy repair attempt.

tommystendahl:
- Validate HTTP status and non-blank response body in JolokiaHttpClient.checkForNotifications(), throwing IOException so existing retry logic handles transient network issues
- Move JSON deserialization into checkForNotificationsWithRetry() retry loop so JacksonException from malformed responses triggers retries instead of bypassing them
- Track consecutive task execution failures per job in ScheduleManagerImpl; after 5 consecutive failures, mark the job as FAILED via new ScheduledJob.markFailed() so it gets purged from the queue instead of looping indefinitely as ON_TIME
- Reset notification failure counters in JolokiaNotificationController when a new repair listener is registered, preventing stale counts from a previous network blip carrying over

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 * Upgrade ecChronos to Spring Boot 4 - Issue #1711
 * Add JSON output format for ecctool state - Issue #1739
+* Fix repair jobs stuck ON_TIME when Jolokia returns empty body due to transient network issues - Issue #1740
 
 ## Version 1.0.6
 
--- a/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/spring/BeanConfigurator.java
+++ b/application/src/main/java/com/ericsson/bss/cassandra/ecchronos/application/spring/BeanConfigurator.java
@@ -33,6 +33,7 @@ import com.ericsson.bss.cassandra.ecchronos.data.sync.EccNodesSync;
 import com.ericsson.bss.cassandra.ecchronos.fm.RepairFaultReporter;
 import com.ericsson.bss.cassandra.ecchronos.utils.exceptions.ConfigurationException;
 import com.ericsson.bss.cassandra.ecchronos.utils.exceptions.EcChronosException;
+import com.ericsson.bss.cassandra.ecchronos.utils.enums.sync.NodeStatus;
 import java.net.InetAddress;
 import java.io.IOException;
 import java.net.UnknownHostException;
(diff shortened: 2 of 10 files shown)


VictorCavichioli (dismissed):
LGTM

jwaeab (approved):

PR #1766, opened 2026-08-28
fix: reuse repair state across replicas

### LINKED ISSUE #1089: Repairs  ranges repeated for each node
Looking at the range repairs done during a VNODE repair I have noticed that it does the repair for all ranges the node contains, not just the primary range. 
So if a keyspace has a replication factor of 3, then each range will be repaired 3 times during a repair cycle of all the nodes on the cluster. This seems like overkill. 

The more normal way to do a repair across a cluster is for primary ranges only, using the -pr option within nodetool, this will ensure each range is repaired only once, per cycle.

One way to do this would just do repairs for primary repairs for each node, however that may cause another inefficiency as described by Tommy:

"For scheduled repair: as long as we make sure a range is only repaired once. If we do that by doing primary ranges only or by making sure each node knows if the range has been repaired by another node doesn't matter that much to me. But I think ecchronos has a feature to combine ranges next to each other which make repair more efficient in large cluster and I think that will be much more useful if don't do only primary."

So further investigation should be made as to which way to fix this is best/practical.

0rlych1kk4:
Fixes #1089 

Repair state was rebuilt using history from only the node currently being evaluated.

That means a successful repair recorded through another replica could be missed, so the same token range could appear unrepaired and be scheduled again.

This updates `calculateNewState()` to read repair history from the replica nodes for the relevant ranges before rebuilding vnode state.

I also added a regression test covering the case where the repair is recorded for one replica and state is calculated for another.

Testing:
- `TestVnodeRepairStateFactoryImpl`: 19/19 passed
- `mvn -pl core.impl -am test` progressed through the reactor, but the `data-agent` Testcontainers tests could not run locally because Docker was not available

### DIFF
--- a/core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/vnode/VnodeRepairStateFactoryImpl.java
+++ b/core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/vnode/VnodeRepairStateFactoryImpl.java
@@ -109,23 +109,39 @@ public class VnodeRepairStateFactoryImpl implements VnodeRepairStateFactory
                 = myReplicationState.getTokenRangeToReplicas(tableReference, node);
         long lastRepairedAt = previousLastRepairedAt(previous, tokenRangeToReplicaMap);
 
-        Iterator<RepairEntry> repairEntryIterator;
+        Set<DriverNode> replicaNodes = new HashSet<>();
+        tokenRangeToReplicaMap.values().forEach(replicaNodes::addAll);
 
-        if (lastRepairedAt == VnodeRepairState.UNREPAIRED)
+        List<RepairEntry> repairEntries = new ArrayList<>();
+
+        for (DriverNode replicaNode : replicaNodes)
         {
-            LOG.debug("No last repaired at found for {}, iterating over all repair entries", tableReference);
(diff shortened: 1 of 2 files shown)


0rlych1kk4:
Hi @tommystendahl  I reproduced the cross-replica repair-state case we discussed in #1089 and added a regression test plus a fix here. When you have time, could you take a look?

tommystendahl:
Hi @0rlych1kk4 .
Thanks for your effort, unfortunately I think you try to solve a problem that dos not really exist.

The repair history is always updated for all nodes involved in the repair, not just the repair coodinator, so when a node reads its repair history it should be up to date.
The locking mechanism used in ecchronos should prevent different nodes from trying to repair the same range in parallell so this should not really happen.
There are probably some corner case where this can happen anyway but this change would be very expensive, the number of queries for repair history quite dramatic. I don't think its worth that to solve a small corner case.

But also want to say that the patch is nice and clean and the tests well structured.

0rlych1kk4:
Thanks @tommystendahl , that makes sense. I was assuming the repair history entry was scoped to the node that initiated the repair, which is why I modeled the regression test that way. If the history is written for every participating node, then the extra replica-wide reads aren't justified.
I'll close this one. Thanks also for the feedback on the patch and tests.

Issue #680, opened 2024-06-17
Create "status" Command on EccTool


VictorCavichioli:
Story Description:
It would be very useful to have a command on ecctool to get the status from the others ecchronos instances, like nodetool status

Acceptance Criteria:

- [ ] Command to observe status of instances working

Definition of Done:

Code designed as described and approved by maintainers

Notes:

NA

pratapmajge:
Hi, I’d like to work on this issue. I’m planning to investigate the existing EccTool commands and implement the status command for observing instance status. If the issue is still available, I’d be happy to take it.

tommystendahl:
Hi @pratapmajge 
I have assigned the issue to you now.

PR #1768, opened 2026-08-28
feat(ecctool): add cluster-wide status command

pratapmajge:
## Summary

Implements #680 by adding cluster-wide status reporting to ecctool.

### Changes

- Added cluster-wide node retrieval through `GET /state/nodes?all=true`
- Updated `ecctool status` to display nodes across all EcChronos instances
- Added `ecctool status --local` for the legacy local health check
- Removed the obsolete `/state/status` endpoint and client method
- Added REST, Python, and integration/Behave test coverage
- Updated ecctool documentation and CHANGES.md

### Testing

- `TestStateManagementRESTImpl`: 3 passed
- `test_ecctool_status.py` + `test_ecctool_state.py`: 11 passed
- `mvn install -DskipTests`: passed
- `git diff --check`: passed

Docker-based integration tests were not run because the Docker integration environment was unavailable.

### Notes

`ecctool status` reports the persisted `nodes_sync` registry and does not guarantee current EcChronos JVM liveness.

### DIFF
--- a/data/src/main/java/com/ericsson/bss/cassandra/ecchronos/data/sync/EccNodesSync.java
+++ b/data/src/main/java/com/ericsson/bss/cassandra/ecchronos/data/sync/EccNodesSync.java
@@ -73,6 +73,7 @@ public final class EccNodesSync
 
     private final PreparedStatement myGetByEccInstanceStatement;
     private final PreparedStatement myGetByEccInstanceAndDCStatement;
+    private final PreparedStatement myGetAllStatement;
     private final Long connectionDelayValue;
     private final ChronoUnit connectionDelayUnit;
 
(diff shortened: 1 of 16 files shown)


data/src/main/java/com/ericsson/bss/cassandra/ecchronos/data/sync/EccNodesSync.java
```diff
@@ -125,6 +126,10 @@ private EccNodesSync(final Builder builder)
                 .build()
                 .setConsistencyLevel(ConsistencyLevel.LOCAL_QUORUM));
 
+        myGetAllStatement = mySession.prepare(selectFrom(KEYSPACE_NAME, TABLE_NAME).all()
```
tommystendahl, line 129: We can't do this. This is a full table scan even if it dosen't require ALLOW FILTERING, this is a to dangerous operation for ecchronos to do in the Cassandra cluster.
Since an ecchronos node dosen't know the identity of other ecchronos nodes we can't read the information node by node either. I don't see any way of solving this.

tommystendahl:
Hi @pratapmajge,
Thank you very much for your effort but I'm afraid we have wasted your time on this, I'm very sorry for that.
There is reliable and safe way to read the data about all nodes from the node_sync table, that's probably way this was never implemented. I should close the issue so no one else wastes time on this.
Otherwise it was a clean and nice patch.

pratapmajge:
Hi Tommy,
Thank you for the clarification and for reviewing the patch. I understand the concern regarding the full table scan and the node identity limitation.
I'll close this out. Thanks again for the feedback and guidance.

Issue #1754, opened 2026-08-20
Race Condition on NodeAddedActions


VictorCavichioli:
During tests with topology changes I noticed the following problem, when a new node is added to a datacenter ecchronos receives the event from the driver, creates the connections, marks node as available in nodes_sync table but does not creates the schedules jobs for this new node, a restart on the pod was needed, I believe that this is because NodeAddedAction calls myDistributedNativeConnectionProvider.addNode(myNode) after NodeWorkerManager.addNode, this might lead to a null on myNode, which makes validateJob(job, myNode) returns false

VictorCavichioli:
I think adding this if should be added on onUp before going to the scheduler, I need to test this

if (myWorkerManager != null)
            {
                myWorkerManager.addNode(node);
            }

0rlych1kk4:
Hi @VictorCavichioli , I was looking into this race condition as well. The ordering between NodeWorkerManager.addNode and myDistributedNativeConnectionProvider.addNode looks important here, especially if validateJob(job, myNode) can run while the worker manager still has no corresponding node state.
Would it be useful if I worked on a regression test that reproduces the topology-change scenario and verifies that schedules are created without requiring a pod restart? I’m happy to help with that if it complements the fix you're testing.

VictorCavichioli:
Hi @0rlych1kk4 !

Yes it would be really helpfull!

0rlych1kk4:
Thanks @VictorCavichioli ! I’ll take a look at the regression test and try to reproduce the node-added scheduling failure on the current code. I’ll focus on making sure the test catches the case where the new node is seen but the schedules are not created until restart.

PR #1769, opened 2026-08-31
test: verify node registration before scheduling

0rlych1kk4:
Adds regression coverage for the node-added scheduling path.

The test verifies that a newly added node is registered with NodeWorkerManager before ScheduleManager.createScheduleFutureForNode() is called.

This protects the ordering required to avoid scheduling a node before worker state has been initialized.

Related to #1754.

Test:
`mvn -pl core.impl -am -Dtest=TestNodeLifecycleHandler -Dsurefire.failIfNoSpecifiedTests=false test`


tommystendahl (approved):

tommystendahl:
@0rlych1kk4 Thanks for your contribution.

PR #1786, opened 2026-09-04
Add Repair History pre and post Job in Incremental path

### LINKED ISSUE #1777: Incremental Repair does not works properly
We noticed the following, incremental repair jobs are being created per node, one repair job per node, which makes ecchronos to do the same work x times more based on the amount of nodes it manages, also, incremental repair jobs are not reading or inserting data in repair_history, which means that the schedule would never now if a repair its on time if the instance is restarted

VictorCavichioli:
Partially closes #1777


tommystendahl (commented):
Reviewed the incremental repair-history changes and the SchemaRefresher refactor. Overall solid with strong test coverage. The per-replica write / single-node read design is correct. A few items to address before merge:

## 🔴 Blocker — Schema migration for `repair_type`

The PR adds a `repair_type text` column to `ecchronos.repair_history`. Existing deployments upgrading to this version will have the table **without** this column. Both the read path (`convertRowToRepairHistoryData` / `RepairEntryIterator` via `row.get(COLUMN_REPAIR_TYPE, ...)`) and the write path (`insertFinish` / `recordCompletedRepair` binding the new column) will fail with "Undefined column repair_type" until the column is added.

Please add:
- An `ALTER TABLE ecchronos.repair_history ADD repair_type text` migration path, and
- An UPGRADE.md entry documenting the required schema change.

Otherwise rolling upgrades break.

## 🟡 RepairGroup — incremental branch sets `myNode` without null-check

```java
else // INCREMENTAL
{
    myRepairHistory = builder.myRepairHistory;
    myNode = builder.myNode;   // no checkNotNull
}
```

The VNODE and PARALLEL_VNODE branches use `checkNotNull(builder.myNode, "Node must be set")`, but the incremental branch does not. In practice `IncrementalRepairJob.iterator()` always supplies a non-null node, but the invariant is no longer enforced at the `RepairGroup` boundary — a null would surface as an NPE deep in `repairHistory.newSession(node, ...)` → `node.getHostId()`. Suggest adding the same `checkNotNull` for consistency and a clearer failure.

## 🟡 Duplicate history rows in multi-instance setups

`recordCompletedRepair` writes one row per replica. In a GeoRed / multi-instance deployment where two instances each manage replicas of the same range, both instances will record rows for the same logical repair, producing overlapping entries. Harmless for correctness (read path takes max `finishedAt` of SUCCESS rows) but it inflates the table. Worth a comment acknowledging this, or dedup consideration.

## 🟡 Fully-qualified type in RepairHistory interface

```java
final java.util.Collection<UUID> replicaNodeIds,
```
Please use an import instead of the fully-qualified name (likely a checkstyle finding).

## ⚪ Minor

- `UNKNOW_REPAIR_TYPE` typo in `RepairHistoryService` — should be `UNKNOWN_REPAIR_TYPE`.
- `lastSuccessfulRunFromHistory` Javadoc says "Reads are keyed by the coordinator node so that repairs performed by any ecChronos instance are visible." The read is keyed by `myNode` (the managed node); it is the per-replica *write* that makes the entry visible, not the read keying. Slightly misleading comment.

Nice work on the test coverage — the peer-completion and per-replica-write tests directly validate the multi-instance behavior, and the NO_OP fallback keeps the legacy `IncrementalRepairTask` path backward compatible.

core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/RepairGroup.java
VictorCavichioli: I think probably now repair history should be mandatory to all repair types

core.impl/src/main/java/com/ericsson/bss/cassandra/ecchronos/core/impl/repair/RepairGroup.java
```diff
@@ -235,12 +240,30 @@ public Collection<RepairTask> getRepairTasks(final UUID nodeID)
...
+            }
+            else
+            {
+                tasks.add(new IncrementalRepairTask(
+                        nodeID,
+                        myJmxProxyFactory,
+                        myTableReference,
+                        myRepairConfiguration,
+                        myTableRepairMetrics));
+            }
```
VictorCavichioli: and this is not needed anymore if repair history is mandatory

VictorCavichioli (commented):
Missing upgrade guide

tommystendahl (commented):
Re-reviewed after the latest commits. All items from the previous review are addressed:

- ✅ **Schema migration** — UPGRADE.md now documents the required `ALTER TABLE ecchronos.repair_history ADD repair_type text;` under an "Upgrade to 1.1.x" section, with a note that it must be applied before upgrading. This resolves the rolling-upgrade blocker.
- ✅ **RepairGroup null-check** — nicely done; rather than just guarding the incremental branch, `myNode` and `myRepairHistory` are now unconditionally required (`checkNotNull`) and `myRepairHistory` is `final`. The `getRepairTasks()` incremental path no longer forks on null and always uses the history-aware task constructor, with `RepairHistory.NO_OP` as the intended no-op fallback. Cleaner than the original suggestion.
- ✅ **Fully-qualified `java.util.Collection`** — now imported.
- ✅ **`UNKNOW_REPAIR_TYPE` typo** — renamed to `UNKNOWN_REPAIR_TYPE`.
- ✅ **Read-keying Javadoc** — corrected to state reads are keyed by `myNode` and visibility comes from the per-replica write.

Compiles cleanly across core, core.impl, data, application.

LGTM.

tommystendahl (dismissed):

tommystendahl (dismissed):

tommystendahl (dismissed):

tommystendahl (approved):

